PR #28, opened 2025-04-22
fix: OneOrManyOne usage fixed

robert-affinidi:


stormer78:
More changes that conflict with this are in the SDK_Websocket_refactor branch. Rejecting this PR as other branch will be released shortly.

PR #31, opened 2025-04-24
fix: workaround for TDKError from DIDPeerError conversion

robert-affinidi:


stormer78:
### Not needed with latest release

PR #122, opened 2025-10-07
feat: Integrate did-cheqd resolver

DaevMithran:


stormer78:
Hello - thanks for the PR.

largely looks good, except the documentation on did-cheqd is the did:example boilderplate.

PR #157, opened 2025-11-28
Added VerificationMethod::get_public_key_bytes() helper method

stormer78:
Converts a multikey DID VerificationMethod to raw public-key bytes


stormer78 (commented):
LGTM

PR #160, opened 2025-12-03
Splitting validate_proof into local and resolve DID methods so you ha…

stormer78:
…ve flexibility on where the public key bytes come from


stormer78 (commented):
LGTM

PR #282, opened 2026-04-14
 chore: harden mediator startup paths and bump version to 0.13.1

durch-affinidi:
## Summary

 This PR hardens mediator startup/runtime code by removing panic-style error paths (unwrap/expect) in
 production code and replacing them with explicit error handling and typed propagation where
 appropriate.

 It also includes a patch version bump for the crate.

 - Crate: affinidi-messaging-mediator
 - Version: 0.13.0 → 0.13.1
 
---

 #### 1) Startup/runtime hardening (server.rs)

 Replaced panic-style startup logic with graceful error handling and logs for:
 - mediator config initialization
 - database initialization
 - streaming task startup
 - TLS cert/key presence and TLS config loading
 - HTTP/HTTPS server bind/serve errors
 - shutdown signal handler installation (Ctrl+C, SIGTERM)

 #### 2) Database initialization hardening (database/initialization.rs)

 Removed expect in initialization paths and mapped failures to typed mediator errors:
 - hardcoded ACL parsing
 - admin account setup calls
 - hardcoded schema migration version parsing

 #### 3) Rate limiter construction hardening

 Updated both:
 - common/rate_limiter.rs
 - common/did_rate_limiter.rs

 to avoid expect on NonZeroU32 and use explicit safe construction logic.

 #### 4) Version bump

 - Cargo.toml patch bump to 0.13.1

### DIFF
--- a/crates/messaging/affinidi-messaging-mediator/CHANGELOG.md
+++ b/crates/messaging/affinidi-messaging-mediator/CHANGELOG.md
@@ -4,6 +4,21 @@
 
 ## 15th April 2026
 
+### 0.13.1 — Startup Hardening
+
+- **BREAKING:** LUA scripts file (`database.functions_file`) is now required in
+  mediator configuration. Previously, omitting it silently skipped script loading
+  and returned successfully; the mediator now exits with an error if it is not
+  specified. Ensure your `mediator.toml` includes a valid `functions_file` path.
+- **CHORE:** Replaced all `unwrap()`/`expect()`/`process::exit()` calls in
+  startup and runtime code with typed `Result` propagation via `MediatorError`
+  - `start()` now returns `Result<(), MediatorError>`; `main()` prints the error
+    and exits non-zero on failure
+  - Database initialization, TLS config, streaming task, DID resolver, and
+    signal handler installation all use explicit error handling
+- **CHORE:** Rate limiter construction uses safe `NonZeroU32` checks instead of
+  `expect()` (applies to both per-IP and per-DID rate limiters)
+
 - **DOC:** Added README section on running without a secure credential store
   (`string://` usage for dev/CI without keyring or AWS Secrets Manager)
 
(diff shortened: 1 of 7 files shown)


stormer78:
Hi @durch-affinidi — thanks for catching this. The intent is exactly right (every `expect()` / `unwrap()` / `process::exit(1)` in startup is a footgun), but this PR has been overtaken by events:

- It's based on `0.13.0`. Main has shipped `0.13.2` since, and the `fix/mediator-deployment` branch (PR #286) sits on top of that with the unified secret backend + wizard rework — `0.13.0 → 0.14.0` covers the same crate.
- `database/initialization.rs::92` (the `Version::parse("0.10.0")` call you patched) was rewritten on the branch into a sequential migration registry, so that hunk no longer applies.
- The branch landed an `error_codes::*` registry of named constants. The raw `12` / `14` / `17` literals in this PR would now be `error_codes::CONFIG_ERROR` / `DB_OPERATION_ERROR` / `INTERNAL_ERROR`.
- The rate-limiter `let-else` you added has a dead branch (`burst.max(1)` is mathematically `≥ 1` for `u32`, so the second `NonZeroU32::new(...)` is `Some` by construction). The intent is right but the cleanest fix is `NonZeroU32::new(burst).unwrap_or(NonZeroU32::MIN)`, which drops the `.max(1)` dance entirely.

Rather than ask you to rebase, I've folded the same hardening into PR #286 directly:

- `server::start` returns `Result<(), MediatorError>`; `main` surfaces the error and exits non-zero (commit https://github.com/affinidi/affinidi-tdk-rs/commit/c6fd5c2).
- `database::initialize` propagates the hardcoded ACL parse + admin setup failures via `?`.
- Rate-limiter constructors use `let-else` + `unwrap_or(NonZeroU32::MIN)` — no `expect()`, no dead branch.
- `shutdown_signal` parks `pending::<()>` on handler-install failure so the surviving signal can still trigger graceful shutdown (kept your idea here verbatim).
- Same breaking change you proposed on missing `database.functions_file` — it's now a hard error, called out in the changelog.

Closing this in favour of #286 with credit. Thanks again for the diagnosis, the original PR was the right read.

PR #298, opened 2026-05-02
chore: release affinidi-tdk-common v0.5.3 — keyring-core + rustls-platform-verifier 0.7

stormer78:
## Summary

- Replaces `keyring 3.x` (single bundled crate) with the new split-out
  `keyring-core 1.0` plus per-target platform store crates
  (`apple-native-keyring-store`, `windows-native-keyring-store`,
  `dbus-secret-service-keyring-store`). Public `secrets` API surface is
  unchanged; a `OnceLock`-guarded `set_default_store(...)` runs lazily on
  first use and no-ops if the host has already registered its own store.
- Bumps `rustls-platform-verifier` 0.6 → 0.7.
- Retires the empty `messaging` Cargo feature in `affinidi-tdk-common`. It
  only gated `use_atm: bool` on the config builder, while the meaningful
  feature gate (with `dep:affinidi-messaging-sdk`) lives on `affinidi-tdk`.
  `use_atm` is now unconditionally present — matching what
  `affinidi-tdk` already does at `tdk/src/lib.rs:145`.

Patch bump only — caret-pinned consumers (`affinidi-tdk = "0.5"`) pick this
up automatically.

## Test plan

- [x] `cargo build -p affinidi-tdk-common` passes
- [x] `cargo check --workspace` passes (all consumers compile)
- [x] `cargo test -p affinidi-tdk-common --no-run` compiles cleanly
- [ ] Manual smoke test: save / load / delete a secret on macOS Keychain
- [ ] Manual smoke test: save / load / delete a secret via Secret Service on Linux
- [ ] Manual smoke test: save / load / delete a secret on Windows Credential Manager

## Out of scope

`affinidi-messaging-mediator` still uses `keyring 3.x` behind its
`vta-keyring` feature (`setup_vta.rs`, `vta_cache.rs`, `helpers.rs`). Same
migration pattern applies — left for a follow-up branch to keep this PR
scoped to the tdk-common 0.5.3 release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,30 @@ Per-crate version history is summarised here; for the full code history see
 
 ## [Unreleased]
 
+## 2026-05-02 — `affinidi-tdk-common` 0.5.3 — keyring-core migration, rustls-platform-verifier 0.7
+
+### Changed
+
+- **`affinidi-tdk-common` 0.5.2 → 0.5.3** — Replaced the bundled `keyring 3.x`
+  dependency with the new split-out
+  [`keyring-core 1.0`](https://crates.io/crates/keyring-core) plus per-target
+  platform store crates: `apple-native-keyring-store` (macOS `keychain`, iOS
+  `protected`), `windows-native-keyring-store`, and
+  `dbus-secret-service-keyring-store` (`crypto-rust`). The public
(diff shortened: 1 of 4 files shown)


stormer78:
Old PR, replaced by 0.6 release

PR #301, opened 2026-05-03
Remove redundant mediator Cargo config

madi-blockstars:
Drop `crates/messaging/affinidi-messaging-mediator/.cargo/config.toml` since the same `--cfg tracing_unstable` rustflag is already set in the workspace root `.cargo/config.toml`. This centralizes build flags and prevents config drift.

### DIFF
--- a/crates/messaging/affinidi-messaging-mediator/.cargo/config.toml
+++ b/crates/messaging/affinidi-messaging-mediator/.cargo/config.toml
@@ -1,2 +1,0 @@
-[build]
-rustflags = ["--cfg", "tracing_unstable"] # needed for object logging


stormer78:
@madi-blockstars Nice catch - thank you for the PR. 
Are you able to resign the commit with a signed signature? we require DCO signed commits

stormer78:
Thanks @madi-blockstars — your suggestion was spot-on. This file was independently removed in #302 (merged 2026-05-04) as part of dropping the `--cfg tracing_unstable` requirement, which addressed exactly the config-drift concern you raised. Closing as already-resolved; credit to you for flagging it.

Issue #324, opened 2026-05-28
unpack(): JWS-in-JWE (sign-then-encrypt) not auto-unwrapped


tarunvadde:
Follow-up from the interop matrix. DIDComm v2.1 §5.3 specifies sign-then-encrypt as the non-repudiation pattern: sign the plaintext into a JWS, then JWE-wrap the JWS bytes. credo-ts's `packSignedAndEncrypted` produces exactly that: outer authcrypt JWE whose plaintext is the JWS.
Affinidi's `unpack()` in `crates/messaging/affinidi-messaging-didcomm/src/message/unpack.rs` decrypts the outer JWE then calls `Message::from_json` on the decrypted bytes directly. For sign-then-encrypt the decrypted bytes are `{payload, signatures}`, not a `Message`, so unpack fails with `"invalid message: missing field id"`.

The underlying primitives are fine. Manual two-step works:
```rust
let decrypted = decrypt(&jwe, &recipient_kid, &recipient_priv, Some(&sender_pub))?;
let inner_jws = std::str::from_utf8(&decrypted.plaintext)?;
let verified = verify_ed25519(inner_jws, &signer_pub)?;
let message = Message::from_json(&verified.payload)?;
```

So it's the high-level dispatcher that needs to recurse: after `decrypt`, parse the bytes as JSON, detect `payload`+`signatures`, run `verify_ed25519`, then `Message::from_json` on the verified payload. The result should probably flow into a richer `UnpackResult::Encrypted `(or a new `EncryptedAndSigned`) that carries `non_repudiation: true` + `signer_kid` alongside the existing fields. That's an API shape change so worth a discussion before patching.

stormer78:
Thanks for the issue report. Confirmed as well that this a bug in the code.

It is an API change, will bump version accordingly for this change.

stormer78:
Addressed in #326

Issue #323, opened 2026-05-28
JWS verify: kid in per-signature unprotected header is ignored


tarunvadde:
Follow-up from the interop matrix (issue #322). credo-ts emits JWS with `kid` in the per-signature unprotected `header` field. RFC 7515 §4.1.4 explicitly permits `kid` in either protected or unprotected per-signature headers, and §7.2.1 (General JSON Serialization) shows the unprotected `header` member alongside `protected`. SICPA didcomm-python actually hard-requires unprotected kid, which is why credo emits it there.

`JwsSignature` in `crates/messaging/affinidi-messaging-didcomm/src/jws/envelope.rs` only declares `protected` and `signature`, so the unprotected `header` member is silently dropped during deserialization. `verify_ed25519` then reads `kid` only from the protected header. The result: credo-packed JWS verifies fine (the signature math is correct), but `UnpackResult::Signed.signer_kid` always comes back as `None`, so callers can't tell who signed.

Suggested fix shape (haven't validated it builds, just sketching the direction):

```diff
@@ src/jws/envelope.rs
 pub struct JwsSignature {
     pub protected: String,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub header: Option<serde_json::Value>,
     pub signature: String,
 }
```
```diff
@@ src/jws/verify.rs (inside verify_ed25519, where signer_kid is built)
-    signer_kid: header.kid,
+    signer_kid: header.kid.or_else(|| {
+        sig_entry.header.as_ref()
+            .and_then(|h| h.get("kid").and_then(|v| v.as_str()).map(String::from))
+    }),
```

Happy to send a PR.

stormer78:
Thanks - confirming this is a bug and will be addressed in the next version of the library.

stormer78:
Addressed in #326

PR #326, opened 2026-05-31
fix: DIDComm v2.1 interop — ECDH-1PU KDF (#322), JWS kid (#323), sign-then-encrypt (#324)

### LINKED ISSUE #322: ECDH-1PU+A256KW concat KDF: cc_tag missing length prefix in SuppPrivInfo
We're testing DIDComm v2.1 interop between `affinidi-messaging-didcomm` and [credo-ts](https://github.com/openwallet-foundation/credo-ts). Authcrypt JWEs don't decrypt either direction.

`concat_kdf_1pu` at `crates/messaging/affinidi-messaging-didcomm/src/crypto/ecdh_1pu.rs:121-122` appends `cc_tag` to the SuppPrivInfo step without a length prefix. NIST SP 800-56A §5.8.2.1.1 length-prefixes every OtherInfo entry, the four preceding entries in the same function already do, and both SICPA didcomm-python (via joserfc) and credo-ts/Askar length-prefix the tag. The derived KEK differs from every spec-following implementation by exactly the four missing bytes.

Verified bidirectionally: with the patch below, credo-packed JWEs decrypt in affinidi and affinidi-packed JWEs decrypt in credo. Without it, both fail with AES-Key-Wrap integrity errors.

```diff
-    // SuppPrivInfo: cc_tag (the auth tag from content encryption)
-    hasher.update(cc_tag);
+    hasher.update((cc_tag.len() as u32).to_be_bytes());
+    hasher.update(cc_tag);
```
@stormer78

stormer78:
Fixes #322, #323, #324 — DIDComm v2.1 interop bugs found via an interop matrix against [credo-ts](https://github.com/openwallet-foundation/credo-ts), verified against the code and the relevant specs. Also hardens the **SDK** consumer path and consolidates duplicated key parsing.

## didcomm crate (the three filed issues)

**#322 — ECDH-1PU Concat KDF missing length prefix on the tag (high).** `cc_tag` was fed to the KDF without the 32-bit big-endian length prefix every other OtherInfo field carries, so the derived KEK differed from every spec-following impl (credo-ts/Askar, didcomm-python) — breaking all `ECDH-1PU+A256KW` authcrypt interop. Confirmed against draft-madden-jose-ecdh-1pu-04 §2.3/App. B. Packing is now spec-correct (X25519/P-256/K-256; anoncrypt unaffected). A transitional **dual-KEK decrypt fallback** (try correct, then legacy on AES-KW failure) keeps upgraded recipients able to read not-yet-upgraded senders; reported via `legacy_kek_used`.

**#323 — unprotected JWS `kid` ignored.** `JwsSignature` gained the per-signature unprotected `header` (RFC 7515 §7.2.1); verify reads protected then unprotected `kid`, so `signer_kid` is populated.

**#324 — sign-then-encrypt not auto-unwrapped.** `unpack()` now detects a JWS inside a decrypted JWE, verifies the inner signature, and reports `non_repudiation` + inner `signer_kid`. `UnpackResult` is now `#[non_exhaustive]`.

## SDK hardening (the real consumer path)

The SDK's `unpack()` had a **parallel** implementation that didn't deliver #323/#324 to clients:
- **Security:** signed JWS were parsed **without verifying the signature**, yet returned with `non_repudiation: true` — a forged signature was accepted and labelled non-repudiable. `unpack()` now resolves the signer's Ed25519 key (via the DID resolver), verifies, and **errors** on an unresolvable signer or bad signature. Signer attributed in `UnpackMetadata.sign_from`.
- **Sign-then-encrypt** is now verified end-to-end through the SDK.

## Consolidation (reduce parallel paths)

Three copies of "DID verification-material → key" parsing collapsed into one: `affinidi-did-common`'s new `VerificationMethod::decode_public_key()` (handles `publicKeyMultibase` + `publicKeyJwk` via `affinidi-crypto`, returns `(multicodec, bytes)`). The SDK sender-key path and DID-authentication's key-agreement resolver both delegate to it. **didcomm stays pure** (no DID-resolution dep); resolution lives in the SDK/auth layers.

## Tests
- didcomm: length-prefix KAT, spec≠legacy KEK, dual-KEK fallback for X25519/P-256/K-256, JWS unprotected-kid + precedence, sign-then-encrypt + missing-key error (98 pass)
- did-common: `decode_public_key` multibase + JWK (Ed25519 / P-256 SEC1) + missing-material
- SDK: JWS verify+attribute, **tampered JWS rejected**, sign-then-encrypt round-trip, all existing authcrypt/anoncrypt round-trips (X25519/P-256/K-256) still green
- Full workspace build + `clippy --all-targets --all-features -D warnings` clean; `cargo deny` shows only pre-existing advisories

## Versions (major.minor pinning)
`affinidi-messaging-didcomm` 0.13.3 → **0.14.0** (interop-affecting). Dependent bumps + didcomm pin `0.13`→`0.14`: sdk 0.18.4, mediator 0.15.7, tdk 0.7.2, helpers 0.13.1, text-client 0.12.3, did-authentication 0.3.4, meeting-place 0.4.1, didcomm-service 0.3.2. Plus **affinidi-did-common 0.3.3 → 0.3.4** (new `decode_public_key` API; patched + `0.3`-pinned, so no further cascade).

## Migration / rollout
0.14 senders emit the spec-correct KEK, so they can't be decrypted by un-upgraded 0.13 recipients. **Upgrade recipients before senders** (the dual-KEK fallback makes upgraded nodes bilingual). No working external interop is lost. The SDK JWS change is stricter: flows that relied on unverified-JWS parsing now get an error.

## Known follow-ups
- **Centralize the crypto core in `affinidi-crypto`** (separate ADR): didcomm hand-rolls its JOSE crypto (ECDH/Concat-KDF/AEAD/key-wrap) — the parallel-crypto root cause of #322. affinidi-crypto today is signing/key-type/JWK/PQC only; growing it + rebasing didcomm is a multi-PR migration.
- **vta-sdk 0.7.0** (external) still pins didcomm `0.13` → pulls the published 0.13.3 (two didcomm versions coexist; `multiple-versions = "allow"`, deny passes). Republish vta-sdk against 0.14, then bump its pin.
- A sibling JS DIDComm lib (e.g. vta-didcomm-js), if it hand-rolls ECDH-1PU, needs the same length-prefix fix to interop with 0.14 senders.

### DIFF
--- a/crates/applications/affinidi-meeting-place/CHANGELOG.md
+++ b/crates/applications/affinidi-meeting-place/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Meeting Place Changelog
 
+## 31st May 2026 (0.4.1)
+
+- Bump `affinidi-messaging-didcomm` to 0.14 (DIDComm v2.1 interop fixes:
+  ECDH-1PU authcrypt KDF #322, JWS unprotected `kid` #323,
+  sign-then-encrypt unpack #324). No Meeting Place API change.
+
 ## 2nd May 2026 (0.4.0)
 
 - **BREAKING:** Migrated to `affinidi-tdk-common` 0.6 — field accesses on
--- a/crates/applications/affinidi-meeting-place/Cargo.toml
+++ b/crates/applications/affinidi-meeting-place/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "affinidi-meeting-place"
-version = "0.4.0"
+version = "0.4.1"
 description = "Affinidi Meeting Place SDK. Discover and connect with others in a secure and private way."
 edition.workspace = true
 authors.workspace = true
@@ -15,7 +15,7 @@ rust-version.workspace = true
 [dependencies]
 affinidi-did-authentication = "0.3"
 affinidi-tdk-common = "0.6"
-affinidi-messaging-didcomm = { path = "../../messaging/affinidi-messaging-didcomm", version = "0.13" }
+affinidi-messaging-didcomm = { path = "../../messaging/affinidi-messaging-didcomm", version = "0.14" }
 affinidi-did-resolver-cache-sdk = "0.8"
 affinidi-did-common = "0.3"
 
--- a/crates/identity/affinidi-did-authentication/CHANGELOG.md
+++ b/crates/identity/affinidi-did-authentication/CHANGELOG.md
@@ -8,6 +8,10 @@
   spec-compliant peers; the 0.14 dual-KEK fallback preserves
   compatibility with not-yet-upgraded peers during rollout. No API
   change in this crate.
+- Internal: the key-agreement key extractor now delegates to
+  `affinidi-did-common`'s `VerificationMethod::decode_public_key`, so the
+  JWK/multibase parsing is shared with the messaging SDK rather than
+  duplicated here.
 
 ## 28th May 2026 (0.3.3)
 
--- a/crates/identity/affinidi-did-authentication/CHANGELOG.md
+++ b/crates/identity/affinidi-did-authentication/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Affinidi DID Authentication
 
+## 31st May 2026 (0.3.4)
+
+- Bump `affinidi-messaging-didcomm` to 0.14. This crate authcrypts the
+  DID-authentication challenge/response (ECDH-1PU), so it directly picks
+  up the corrected authcrypt KDF (#322) and is now interoperable with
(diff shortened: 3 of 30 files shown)


stormer78:
Tracking issues opened for the two follow-ups noted above:
- #327 — centralize the JOSE crypto core in `affinidi-crypto` and rebase didcomm onto it (the parallel-crypto root cause of #322).
- #328 — republish `vta-sdk` against didcomm 0.14 and bump its pin (removes the duplicate 0.13.3 in the tree).

PR #355, opened 2026-06-05
fix: select recipient key-agreement key matching sender curve

marat-affinidi:
When packing authcrypt messages, the sender and recipient must use key-agreement keys on the same elliptic curve. Previously, both `_pack_encrypted_for_did` (did-authentication) and `pack_encrypted` (messaging-sdk) blindly picked the first key-agreement key from the recipient's DID document. If the recipient advertised multiple curves (e.g. secp256k1, P-256, X25519) and the first didn't match the sender's curve, ECDH-1PU failed with "curve mismatch between private and public keys".

Fix: resolve the sender's curve first, then iterate the recipient's key-agreement keys and pick the first one on the matching curve.

Reproduces with any X25519 sender (did:key or did:webvh) talking to a mediator whose first keyAgreement entry is secp256k1 or P-256.

### DIFF
--- a/crates/identity/affinidi-did-authentication/src/lib.rs
+++ b/crates/identity/affinidi-did-authentication/src/lib.rs
@@ -687,6 +687,11 @@ where
 ///
 /// Resolves both DIDs to get key agreement keys, looks up the sender's
 /// private key from the secrets resolver, and calls authcrypt.
+///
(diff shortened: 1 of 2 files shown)


stormer78:
Nice work @marat-affinidi in finding this...

Issue #350, opened 2026-06-04
mediator-setup: file:// secret-backend URL with relative path is RFC-malformed — silently writes outside working dir, fails for non-root users


geoffturk:
### Summary

When the mediator-setup wizard constructs a file:// URL from a relative path for the secret backend, the resulting URL is malformed per RFC 3986 and either:

- Runs as root → silently writes the secrets file to the filesystem root (/) instead of the wizard's working directory. The setup appears to succeed but the file lands somewhere the user didn't expect.
- Runs as any non-root user → fails the backend probe with a confusing permission-denied error.

###  Reproduction

mediator-setup version 0.15.6 on Ubuntu 26.04.

1. cd ~/mediator && mediator-setup
2. Step through the wizard. At Key Storage, choose Local file (file://).
3. When prompted for "Storage file path", enter a relative path (e.g. conf/secrets.json) — or accept whatever default the wizard offers if it's relative.
4. Continue to the end of the wizard.

###  Observed

As a non-root user, the wizard aborts on the final provisioning step:

```err
Provisioning unified secret backend: file://conf/secrets.json
Error: Secret backend 'file://conf/secrets.json' failed probe: I/O error on backend 'file': Permission denied (os error 13)
```

Verified via strace and a direct touch /secrets.json test (also EACCES for non-root): the binary is opening `/secrets.json` at the filesystem root, not `./conf/secrets.json` in the working directory. This is correct RFC 3986 behavior — `file://conf/secrets.json` parses to `authority=conf, path=/secrets.json` — but it's almost certainly not what the wizard intends to express.

As root, the same wizard run succeeds, but the secrets file is actually written to /secrets.json rather than ~/mediator/conf/secrets.json. The user is unlikely to notice.

###  Expected

When the user supplies a relative path, the wizard should construct a URL that resolves to a path inside the working directory — either by resolving to an absolute path at prompt time (file:///<cwd>/conf/secrets.json, three slashes) or by rejecting the relative form with a clear error explaining the three-slash requirement.

###  Workaround

  Enter the path as an absolute URL:

  /home/<user>/mediator/conf/secrets.json

###  Suggested fix

  In whatever code constructs the backend URL from the wizard's "Storage file path" prompt:

  - If the user-entered path is relative, canonicalize against std::env::current_dir() before formatting as a URL.
  - Always emit file:///<absolute path> (three slashes / empty authority).
[...]

PR #356, opened 2026-06-06
fix: mediator-setup no longer clobbers unified secret backend (#354)

### LINKED ISSUE #354: mediator-setup` clobbers unified secret backend when `[secrets].storage = "file:///…/secrets.json"`
**Affected versions:** Mediator 0.15.13 (upstream `main` as of 2026-06-05), likely earlier as well.

  ### Summary

  When `[secrets].storage` is a `file://` URL whose path's filename is `secrets.json`, the wizard's unified-backend writes are silently overwritten with a legacy `affinidi_secrets_resolver`-format array. The mediator binary then can't find its operating secrets / admin credential and refuses to start with `Configuration Error: No operating secrets found`. On each crash-restart the file backend's open path re-initialises the file to `{"entries": {}}`, masking the original wizard output.

  ### Reproduction

  ```toml
  # recipe.toml
  [deployment]
  type      = "server"
  protocols = ["didcomm"]
  use_vta   = true
  vta_mode  = "sealed-export"

  [vta]
  context = "mediator"

  [secrets]
  storage = "file:///var/lib/mediator-svc/conf/secrets.json"

  # ... (typical server recipe)
  [output]
  config_path    = "/var/lib/mediator-svc/conf/mediator.toml"
  listen_address = "0.0.0.0:7037"
  ```

  Run Phase 1 + Phase 2 of a sealed-export handoff. Inspect `secrets.json` after Phase 2 succeeds — it's a `[{id, type, privateKeyJwk}, …]` array, not the expected `{"entries": {"mediator_admin_credential": …, …}}` envelope shape. The wizard's own log lines (`✔ mediator_admin_credential`, `✔ mediator_operating_secrets (5 keys)`, etc.) suggested everything was written correctly.

  ### Root cause

  `crates/messaging/affinidi-messaging-mediator/tools/mediator-setup/src/config_writer.rs:76-83`:

  ```rust
  if config.secret_storage == STORAGE_FILE {
      let secrets_path = config_dir(config).join("secrets.json");
      crate::generators::secrets::write_secrets_file(
          &generated.mediator_secrets,
          &secrets_path.to_string_lossy(),
      )?;
  }
  ```

  This block runs after the unified-backend writes and hard-codes the filename to `secrets.json` under `config_dir`. `generators::secrets::write_secrets_file` writes the legacy resolver-format array — and clobbers the unified backend file whenever the operator's `[secrets].storage` URL points at the same path (the natural choice for `file://` backends).

  The doc-string in `generators/secrets.rs` ("The mediator expects a JSON array of Secret objects.") is also stale.

  ### Expected

  Either:

  - The legacy `write_secrets_file` block is removed (the unified backend handles persistence), or
  - It writes to a distinct filename so it can't collide with `[secrets].storage`, or
  - It's gated so it only runs when the unified backend write target is different from `<config_dir>/secrets.json`.

  ### Workaround

  Point `[secrets].storage` at a filename other than `secrets.json`, e.g. `file:///var/lib/mediator-svc/conf/mediator-secrets.json`. The legacy writer still produces a stray `secrets.json` (now inert), but the unified backend file under the new name survives intact.

  ### Impact

  Any operator running a sealed-export or sealed-mint setup against a `file://` secret backend at the conventional `<config_dir>/secrets.json` path. The failure mode is a crash-loop on `systemctl start mediator-svc` that re-initialises the file backend  to `{"entries": {}}` before the operator can read what was originally there — so the symptom looks like "the wizard never wrote anything" rather than "the legacy writer clobbered the wizard's writes."

stormer78:
Fixes #354.

## Problem

`mediator-setup` provisions secrets in two phases:

1. `provision_secret_backend` opens the unified secret backend at the operator's `[secrets].storage` URL and writes the `{"entries": …}` envelope (JWT secret, operating secrets, admin credential).
2. `write_config` (run immediately after) contained a **legacy** block that wrote an `affinidi_secrets_resolver`-format *array* to a **hard-coded** `<config_dir>/secrets.json`.

When the storage path resolved to that same file — which the default `conf/secrets.json` (`DEFAULT_SECRET_FILE_PATH`) always does — phase 2 **clobbered** the envelope phase 1 had just written. The mediator then found no operating secrets / admin credential and crash-looped on start with `Configuration Error: No operating secrets found`, re-initialising the file to `{"entries": {}}` on each restart — so the symptom looked like "the wizard never wrote anything."

## Root cause

The legacy writer predates the unified secret backend. Nothing reads its array format anymore — the running mediator loads operating secrets from the unified store's `OPERATING_SECRETS` entry, never from a raw array file.

## Fix

Remove the legacy write entirely, along with the now-dead `GeneratedValues.mediator_secrets` field and the orphaned `generators::secrets` module. The unified backend is the sole owner of secret persistence.

**Runtime save-back is unaffected** — it was always handled by the `file://` backend's own `put`/`delete` → `save_locked` path, not by this one-shot wizard block. This is a pure deletion with no path-handling logic added (so no risk of re-introducing the kind of `file://` parsing bug fixed in #350).

## Tests

- **New:** `generate_and_write_file_backend_on_default_path_survives` — points the `file://` backend at the default `<config_dir>/secrets.json` (the exact collision) and asserts the unified `{"entries": …}` envelope survives with the well-known keys present.
- **Updated:** the existing end-to-end test previously *asserted* the legacy array was written; it now asserts it is **not**.

## Verification

- `cargo build -p affinidi-messaging-mediator-setup` ✅
- `cargo test -p affinidi-messaging-mediator-setup` ✅ (342 passed)
- `cargo clippy -p affinidi-messaging-mediator-setup --all-targets` ✅ (clean)
- `cargo fmt -- --check` ✅

Bumps `affinidi-messaging-mediator-setup` to 0.1.7 and adds a CHANGELOG entry. (`publish = false` binary — no dependent crates to bump.)


stormer78:
Follow-up commit `91c813e`: also fixed the two summary banners (`print_completion_summary`, `print_final_summary`) that reported the secrets file as a hard-coded `conf/secrets.json` regardless of the operator's `[secrets].storage` path. They now print `config.secret_file_path`, so a `file://` backend at a custom location is reported correctly. Default path unchanged; both sites are already gated on the file:// backend.

Folded into the same (unpublished) **0.1.7** entry so it's a single crate publish. Re-verified: build ✅, 342 tests ✅, clippy clean ✅, `fmt --check` ✅.

PR #380, opened 2026-06-07
fix: reproducible cross-mediator DIDComm forwarding (4 fixes)

maratsh:
## Summary

Four related fixes that together make **cross-mediator** DIDComm v2 message relay work
end-to-end and **reproducibly** between two independent mediators, plus a security gate so the
relax of inbound auth only applies to mediators explicitly configured as relays. Discovered while
standing up two local `did:peer` mediators (ports 7037/7038, `redis-backend`) and running the
`cross_mediator_forwarding` example.

Before this series the example could not complete a cross-mediator round-trip; after the first
three fixes it round-tripped **once** against a cold Redis but failed on every subsequent run — the
4th fix closes that regression.

## The fixes

**1. `cross_mediator_forwarding` example — double-wrap the forward envelope**
`send_message()` built a single forward encrypted *for the recipient's mediator* and sent it *to the
sender's mediator*, which couldn't decrypt it and fell through to the direct-delivery error path.
Correct pattern: `inner = forward(packed, target=RecipientMediator, next=Recipient)` then
`outer = forward(inner, target=SenderMediator, next=RecipientMediator)`.

**2. `/inbound` rejected anonymous requests — blocked inter-mediator relay**
`ForwardingProcessor` POSTs to the remote mediator with no JWT (the inner forward's authcrypt *is*
the authentication), but the `Session` extractor returned `401` first. Added a `MaybeSession`
extractor: Bearer → authenticated session; no Bearer → anonymous session with `global_acl_default`
**— but only when the mediator is configured as a relay (see fix #5)**.

**3. `force_session_did_match` broke anonymous sessions**
The sender-DID check compared the JWE sender against `session.did = ""` for anonymous sessions and
always failed. Now gated on `session.authenticated`.

**4. Forwarding sender never registered — relay not reproducible** (the regression)
In `routing.rs::forward_message`, the `next_account` (recipient) branch persists the next hop via
`account_add(global_acl_default)` when missing, but the `from_account` (sender) branch only built an
*in-memory* `Account` and never persisted it. `store_message`'s Redis Lua then `HINCRBY`s queue
counters onto `DID:<sender>`, creating a row with **only** `SEND_QUEUE_*` fields — no `ACLS`, never
added to `KNOWN_DIDS`. On the next forward, `account_get` returns that phantom row with `acls = 0`
(DENY_ALL), so the `SEND_FORWARDED` gate rejects the relay with `403 e.p.authorization.send_forwarded`.
Fix: mirror the `next_account` branch — `account_add` the forwarding sender from `global_acl_default`
on first contact.

**5. Gate anonymous `/inbound` on relay capability (`SEND_FORWARDED`)**
Fix #2 alone moved the inbound auth boundary for **every** mediator: with the shipped secure default
(`DENY_ALL,LOCAL,SEND_MESSAGES,RECEIVE_MESSAGES`) an anonymous caller still gained
`SEND_MESSAGES`/`RECEIVE_MESSAGES`/`LOCAL` — strictly more than the previous `401`. `MaybeSession`
now synthesizes the anonymous session **only when `global_acl_default` grants `SEND_FORWARDED`** —
i.e. the operator has opted into acting as an inter-mediator relay (capability re-checked per-message
in `routing.rs`). Otherwise the original `AuthError` is propagated, so a mediator on the shipped
default keeps rejecting unauthenticated `/inbound` exactly as before. `ALLOW_ALL` enables relay (as
the cross-mediator example already requires); the secure default does not. Uses capability-bit
checking to match the existing `routing.rs` ACL gates rather than whole-set equality against
`ALLOW_ALL`. Unit-tested in `jwt_auth.rs`.

## Reproduction / verification

Two `did:peer` mediators on :7037 / :7038, `--no-default-features --features didcomm,redis-backend`
(the `ForwardingProcessor` is `#[cfg(feature = "redis-backend")]`-only), with
`global_acl_default = "ALLOW_ALL"`, then:

```
cargo run -p affinidi-messaging-helpers --example cross_mediator_forwarding -- \
  --alice-environment alice --bob-environment bob
```

- Before #4: 1st run (cold Redis) delivers both ways; every later run →
  `403 e.p.authorization.send_forwarded`.
- After #4: from a clean Redis, **5 consecutive runs** deliver both directions (~40ms localhost).
- After #5: on a mediator without `SEND_FORWARDED` in `global_acl_default`, unauthenticated
  `/inbound` is rejected (`401`) rather than silently treated as anonymous.

Tests: `affinidi-messaging-mediator` lib (101, incl. 2 new gate tests) and integration
`test_sender_authentication` (9), 0 failures, with `--features redis-backend`.

## Security / trust note

With this series, anonymous inbound relay is **gated behind explicit relay configuration**
(`global_acl_default` granting `SEND_FORWARDED`, i.e. `ALLOW_ALL` in practice). A mediator left on
the shipped secure default is unaffected and keeps rejecting unauthenticated `/inbound`. Operators
who enable relay accept an **open-relay** posture between trusted mediators; for stricter production
use, gate further on a per-DID ACL or an allowlist of the inner forward's authcrypt sender.

### DIFF
--- a/crates/messaging/affinidi-messaging-helpers/examples/cross_mediator_forwarding.rs
+++ b/crates/messaging/affinidi-messaging-helpers/examples/cross_mediator_forwarding.rs
@@ -152,6 +152,7 @@ async fn main() -> Result<(), ATMError> {
     let msg_id = send_message(
         &alice_atm,
         &atm_alice,
+        &alice_mediator_did,
         &atm_bob,
         &bob_mediator_did,
         msg_text,
@@ -196,6 +197,7 @@ async fn main() -> Result<(), ATMError> {
     let response_id = send_message(
         &bob_atm,
         &atm_bob,
+        &bob_mediator_did,
         &atm_alice,
         &alice_mediator_did,
         response_text,
@@ -261,6 +263,7 @@ async fn main() -> Result<(), ATMError> {
             let ping_id = send_message(
                 &alice_atm,
                 &atm_alice,
+                &alice_mediator_did,
                 &atm_bob,
                 &bob_mediator_did,
                 &ping_msg,
@@ -284,6 +287,7 @@ async fn main() -> Result<(), ATMError> {
             let pong_id = send_message(
                 &bob_atm,
                 &atm_bob,
+                &bob_mediator_did,
                 &atm_alice,
                 &alice_mediator_did,
                 &pong_msg,
@@ -338,10 +342,16 @@ async fn main() -> Result<(), ATMError> {
     Ok(())
 }
 
-/// Send a message from sender to recipient through the recipient's mediator
+/// Send a message from sender to recipient, routing through both mediators when they differ.
+///
+/// When sender and recipient use different mediators, two forward layers are required:
+///   outer (encrypted for sender_mediator, next=recipient_mediator) wraps
+///   inner (encrypted for recipient_mediator, next=recipient).
+/// This ensures each mediator only decrypts its own layer.
 async fn send_message(
     sender_atm: &affinidi_messaging_sdk::ATM,
     sender_profile: &std::sync::Arc<ATMProfile>,
+    sender_mediator_did: &str,
     recipient_profile: &std::sync::Arc<ATMProfile>,
     recipient_mediator_did: &str,
     body_text: &str,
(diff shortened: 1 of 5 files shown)


marat-affinidi:
Closing in favor of a new PR from upstream branch (fork checkout doesn't work with reusable pipeline)

PR #382, opened 2026-06-08
fix: bbs-2023 safe-mode expansion — reject forged undefined attributes (#381)

### LINKED ISSUE #381: Soundness: bbs_2023_transform accepts forged disclosed attribute values (holder tampers base + re-derives)
## Summary

`affinidi-data-integrity` **0.7.4**'s `bbs_2023_transform` selective-disclosure cryptosuite is **unsound**: a credential **holder** can change the value of a signed attribute in their base-proof VC, derive a disclosure proof over the tampered value, and `verify_derived_proof` **accepts it**. This lets a holder present arbitrary, never-issued attribute values that pass verification.

The underlying `affinidi-bbs` 0.3.0 primitive is **sound** (it rejects the same tampered disclosure), so the defect is in the **document/cryptosuite layer** (`bbs_2023_transform`), not the BBS core.

This affects both the basic suite and, by extension, the pseudonym (`0xd95d08`/`0xd95d09`) layer built on the same message handling.

## Threat model

BBS selective disclosure must resist exactly this: the holder controls their own base credential and can re-derive at will. Soundness requires that a derived proof verify **only** for the attribute values the issuer actually signed.

## Reproduction — document layer (BUG)

```rust
use affinidi_bbs as bbs;
use affinidi_data_integrity::bbs_2023_transform as tx;
use serde_json::json;

let sk = bbs::keygen(b"key-material-at-least-32-bytes!!", b"").unwrap();
let pk = bbs::sk_to_pk(&sk);
let did = affinidi_crypto::bls12381::g2_pub_to_did_key(&pk.to_bytes());
let vc = json!({
    "@context": ["https://www.w3.org/ns/credentials/v2"],
    "type": ["VerifiableCredential", "MembershipCredential"],
    "issuer": did,
    "credentialSubject": { "id": "did:key:zMember", "memberLevel": "gold" }
});
let mandatory = ["/@context", "/type", "/issuer", "/credentialSubject/id"];
let base = tx::sign_base_document(
    &vc, &mandatory, &format!("{did}#bbs-key-0"),
    "2020-01-01T00:00:00Z", &sk, &pk, b"32-byte-hmac-key-padding-xxxxxxx",
).unwrap();

// Holder tampers a SIGNED, non-mandatory attribute, then derives a proof
// disclosing only that (forged) attribute:
let mut tampered = base.clone();
tampered["credentialSubject"]["memberLevel"] = json!("platinum");
let derived = tx::create_derived_proof(
    &tampered, &["/credentialSubject/memberLevel"], b"nonce", &pk,
).unwrap();

assert_eq!(derived["credentialSubject"]["memberLevel"], "platinum");
// EXPECTED: false (forged value).  ACTUAL: true  <-- soundness break
assert_eq!(tx::verify_derived_proof(&derived, &pk).unwrap(), true);
```

Same result with full disclosure (`&[""]`).

## Reproduction — raw `affinidi-bbs` (CORRECT, for contrast)

```rust
use affinidi_bbs as bbs;
let sk = bbs::keygen(b"key-material-at-least-32-bytes!!", b"").unwrap();
let pk = bbs::sk_to_pk(&sk);
let sig = bbs::sign(&sk, &pk, b"hdr", &[b"m0".as_ref(), b"gold"]).unwrap();

// Prove disclosure of index 1 but with a tampered message:
let proof = bbs::proof_gen(&pk, &sig, b"hdr", b"ph", &[b"m0".as_ref(), b"platinum"], &[1]).unwrap();
// Correctly rejects:
assert_eq!(bbs::proof_verify(&pk, &proof, b"hdr", b"ph", &[b"platinum".as_ref()], &[1]).unwrap(), false);
```

## Expected vs actual

- **Expected:** `verify_derived_proof` returns `false` (or `create_derived_proof` errors) when the disclosed statements do not match the issuer-signed messages.
- **Actual:** `verify_derived_proof` returns `true` for forged disclosed values.

## Suspected area

The disclosed-statement ↔ BBS-message binding in `bbs_2023_transform::create_derived_proof` / `verify_derived_proof` (the `adj_selective` / `selective_indexes` mapping and/or the disclosed-message reconstruction in verify). The raw `proof_gen`/`proof_verify` enforce the binding; the document layer appears to lose it when re-deriving from a mutated base document.

## Versions

- `affinidi-data-integrity` 0.7.4
- `affinidi-bbs` 0.3.0
- `affinidi-rdf-encoding` 0.1.2

## Impact / downstream

Any vc-di-bbs verifier built on this layer accepts holder-forged attribute values. Blocks OpenVTC/verifiable-trust-infrastructure#295 (bbs-2023 holder-binding) and makes the basic bbs-2023 verifier path unsafe to ship. Strongly reinforces that the `affinidi-bbs` / `bbs_2023` security audit must cover the **document cryptosuite layer**, not just the curve primitives.

stormer78:
Closes #381.

## Summary

The `bbs-2023` selective-disclosure cryptosuite was **unsound**: a credential holder could change the value of an attribute that the credential's `@context` does **not** define (e.g. `memberLevel` under the bare `credentials/v2` context), re-derive a fresh disclosure proof, and have `verify_derived_proof` **accept** the forged value.

## Root cause

JSON-LD expansion **silently dropped unmapped terms** (`affinidi-rdf-encoding/src/jsonld/expand.rs` — `_ => continue`). An undefined attribute therefore produced **no RDF statement at all**: it was neither covered by the issuer's signature nor checked by the verifier, yet it remained in the JSON envelope that applications read. Instrumentation on the PoC showed `non_mandatory_indexes = []` — the BBS proof was over an empty message set, so it verified regardless of the JSON value.

The `affinidi-bbs` primitive is sound (it rejects the equivalent raw tamper), and the suspected `adj_selective` / index mapping was actually correct — the message set simply never contained the term. This is the W3C Data Integrity **"safe mode"** gap.

## Fix

Enforce JSON-LD safe mode on every sign / derive / verify path.

### `affinidi-rdf-encoding` 0.1.4 → 0.1.5

Add `expand_document_safe` / `expand_and_to_rdf_safe`, which error (naming the offending input) on **anything lenient expansion would silently drop**. All three drop sites are covered:

- a **property key** not defined by the active `@context`;
- a node **`@type`** that doesn't resolve to an absolute IRI (no term definition, no `@vocab`) — rejected rather than passed through as a relative IRI; non-string `@type` entries are rejected too;
- a free-standing **scalar** where a node is expected (document root or a node array).

The lenient `expand_document` / `expand_and_to_rdf` defaults are **unchanged** — backward-compatible, additive API.

### `affinidi-data-integrity` 0.7.4 → 0.7.5

`sign_base_document` (issuer), the derive path (`canonicalize_hmac` / `to_deskolemized_nquads`), and `verify_derived_proof` / `verify_pseudonym_derived_proof` (verifier) now canonicalize via the safe variant. (`proof_hash` stays lenient — it processes proof options, not claims.)

- Issuers **refuse** to sign a credential with an undefined claim.
- Verifiers **reject** a reveal document carrying an undefined term.
- Both the basic and pseudonym (`0xd95d08`/`0xd95d09`) suites are covered.

Credentials whose claims are all defined by their context (e.g. via `@vocab`) are unaffected — all W3C `vc-di-bbs` KAT vectors still pass byte-for-byte (13/13 transform + full suites green).

## Tests added

- **(A) Adversarial re-derivation harness** — encodes the threat model (holder owns the base credential and re-derives at will): for every leaf claim, tamper-in-base then re-derive must never verify, with an honest-derivation control so it can't pass vacuously.
- **(B) Coverage/completeness** — issuer rejects an undefined claim at issuance; verifier rejects an undefined term in a reveal document.
- **(C) `affinidi-rdf-encoding` safe-mode unit tests** — reject unmapped property, nested-unmapped term, unmapped `@type`, top-level scalar, and stray-scalar-in-array; accept fully-mapped (`@vocab`) docs, custom types via `@vocab`, and standard VC terms; plus a lenient-drop guard pinning the old behaviour.

## Why we missed it originally

All conformance fixtures (`windDoc` with `@vocab`, `license.json`) are fully mapped, so the drop could never trigger; the two existing negative tests mutate the reveal **after** derivation (keeping the frozen proof bytes) rather than re-deriving as an attacker would. The new tests close both gaps.

## Verification

- `cargo test` — both crates green (incl. all `vc-di-bbs` KATs)
- `cargo clippy --all-targets --all-features` — clean
- `cargo fmt --all -- --check` — clean
- `cargo deny check` — advisories/bans/licenses/sources ok
- downstream `cargo check -p affinidi-vc --all-features` — clean

## Commits

1. `fix: affinidi-data-integrity 0.7.5 — bbs-2023 safe-mode expansion rejects forged undefined attributes (#381)` — core soundness fix + safe-mode for unmapped property keys.
2. `fix: affinidi-rdf-encoding safe mode — also reject unmapped @type and dropped scalars (#381)` — extend safe mode to the remaining two drop sites.

### DIFF
--- a/crates/credentials/affinidi-data-integrity/CHANGELOG.md
+++ b/crates/credentials/affinidi-data-integrity/CHANGELOG.md
@@ -1,5 +1,35 @@
 # Affinidi Data Integrity Changelog
 
+## 8th June 2026 Release 0.7.5
+
+### Fixed
+
+- **`bbs-2023` soundness — forged undisclosed/undefined attribute values
+  (issue #381).** A credential holder could change the value of an attribute
+  that the credential's `@context` does **not** define (e.g. `memberLevel`
+  under the bare `credentials/v2` context), re-derive a fresh disclosure proof,
+  and have `verify_derived_proof` accept it. Root cause: JSON-LD expansion
+  **silently dropped** unmapped terms, so such an attribute was never part of
+  the signed RDF dataset — neither covered by the issuer's signature nor checked
+  by the verifier — yet it remained in the JSON envelope that applications read.
+  The underlying `affinidi-bbs` primitive was sound; the defect was in the
+  document/cryptosuite layer. Fix: every sign / derive / verify path now
+  canonicalizes via JSON-LD **safe-mode** expansion
+  (`affinidi-rdf-encoding::jsonld::expand_and_to_rdf_safe`), which errors on any
+  term not defined by the active `@context`. Consequences:
+  - Issuers (`sign_base_document`) **refuse** to sign a credential with an
+    undefined claim instead of emitting one with an unprotected attribute.
+  - Verifiers (`verify_derived_proof` / `verify_pseudonym_derived_proof`)
+    **reject** a reveal document carrying an undefined term.
+
+  Applies to both the basic and pseudonym (`0xd95d08` / `0xd95d09`) suites.
+  Credentials whose claims are all defined by their context (e.g. via `@vocab`)
+  are unaffected — all W3C `vc-di-bbs` KAT vectors still pass byte-for-byte.
+
+### Changed
+
+- Bump `affinidi-rdf-encoding` to `0.1.5` (adds safe-mode expansion).
+
 ## 7th June 2026 Release 0.7.4
 
 ### Changed
--- a/crates/credentials/affinidi-data-integrity/Cargo.toml
+++ b/crates/credentials/affinidi-data-integrity/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "affinidi-data-integrity"
 description = "W3C Data Integrity Implementation"
-version = "0.7.4"
+version = "0.7.5"
 edition.workspace = true
 authors.workspace = true
 readme = "README.md"
--- a/crates/credentials/affinidi-data-integrity/src/bbs_2023_transform.rs
+++ b/crates/credentials/affinidi-data-integrity/src/bbs_2023_transform.rs
@@ -364,11 +364,14 @@ pub fn verify_derived_proof(
     let proof_hash = proof_hash(&proof_config)?;
 
     // Canonicalize the reveal document (minus proof), relabel via the label map.
+    // Safe mode: reject a reveal document carrying any term not defined by its
+    // @context — such a term would be dropped from the RDF and thus escape the
+    // BBS check while still appearing in the JSON (issue #381).
     let mut doc = reveal_document.clone();
     doc.as_object_mut()
         .ok_or_else(|| malformed("document must be an object"))?
         .remove("proof");
-    let dataset = jsonld::expand_and_to_rdf(&doc).map_err(canon_err)?;
+    let dataset = jsonld::expand_and_to_rdf_safe(&doc).map_err(canon_err)?;
     let (canonical_c14n, _) = rdfc1::canonicalize_with_label_map(&dataset).map_err(canon_err)?;
     let labeled = lines_with_newline(&relabel_and_sort(&canonical_c14n, &parsed.label_map));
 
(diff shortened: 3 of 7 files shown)


stormer78:
### Follow-up: safe mode now covers all three JSON-LD drop sites

Extended `expand_*_safe` beyond unmapped property keys to also reject the two remaining ways data could be silently discarded during canonicalization:

- **Unmapped `@type`** — a node type that doesn't resolve to an absolute IRI (no term definition, no `@vocab`) now errors instead of being passed through as a relative IRI; non-string `@type` entries are rejected too.
- **Dropped scalars** — a free-standing scalar where a node is expected (document root or a node array) now errors instead of being dropped.

Lenient defaults unchanged; all W3C `vc-di-bbs` KAT vectors still pass byte-for-byte. Added four safe-mode unit tests (unmapped-type reject, custom-type-via-`@vocab` accept, top-level scalar reject, stray-scalar-in-array reject). `clippy`/`fmt`/`deny` clean.

PR #388, opened 2026-06-09
feat: per-hop re-wrapping for inter-mediator relay (relay_mode=rewrap) (#385 item 3)

### LINKED ISSUE #385: Harden mediator-to-mediator (relay) routing: scope anonymous inbound, tighten auth fallthrough, add per-DID relay trust
## Background

PR #383 (`fix: reproducible cross-mediator DIDComm forwarding`) made cross-mediator DIDComm v2 relay work end-to-end and reproducibly, and correctly keeps the **shipped secure default safe**: a mediator whose `global_acl_default` lacks `SEND_FORWARDED` continues to reject unauthenticated `/inbound`. Enabling relay is opt-in (the operator must grant `SEND_FORWARDED` in `global_acl_default`, i.e. `ALLOW_ALL` in practice).

That PR is intentionally scoped as a dev/PoC relay between two trusted mediators and documents the resulting **open-relay** posture. This issue captures the follow-on work to make that posture safe for production use, plus general relay-path robustness.

Reference: PR #383 (`fix/cross-mediator-forwarding`), files:
- `crates/messaging/affinidi-messaging-mediator/src/common/jwt_auth.rs` (`MaybeSession`, `anonymous_inbound_allowed`)
- `crates/messaging/affinidi-messaging-mediator/src/handlers/message_inbound.rs`
- `crates/messaging/affinidi-messaging-mediator/src/messages/protocols/routing.rs`

## Motivation

When relay is enabled, an unauthenticated caller currently obtains a full anonymous session with `acls = global_acl_default` (ALLOW_ALL) and `account_type = Standard`, which flows into the **entire** `/inbound` DIDComm dispatch — not only the `forward` relay path. Admin protocols remain role-gated and direct-delivery checks the envelope sender's ACLs, so the practical exposure is limited, but the boundary is broader than "relay forwarding" and trust between relaying mediators is currently all-or-nothing.

### 1. Tighten the auth fallthrough in `MaybeSession` (correctness / hardening)
`MaybeSession::from_request_parts` currently catches **every** `AuthError` and downgrades to an anonymous session when relay is enabled. That includes `InvalidToken`, `ExpiredToken`, `Blocked`, and `InternalServerError`.

- Only synthesize the anonymous session for the genuinely-no-credentials case (`AuthError::MissingCredentials`); propagate all other errors.
- In particular, an `InternalServerError` (e.g. `get_session` fails because the backend is down) should surface as 500, not silently proceed as anonymous.

```rust
match Session::from_request_parts(parts, state).await {
    Ok(session) => Ok(MaybeSession(session)),
    Err(AuthError::MissingCredentials) if anonymous_inbound_allowed(..) => Ok(MaybeSession(anon)),
    Err(e) => Err(e),
}
```

### 2. Scope the anonymous session to the relay capability only (least privilege)
Instead of handing the anonymous session the full `global_acl_default` (ALLOW_ALL), synthesize it with a **minimal** ACL set — just the forward capabilities (`SEND_FORWARDED` / `RECEIVE_FORWARDED`) — so anonymous inter-mediator callers get exactly the relay capability and nothing else. This closes the broader `/inbound` ACL surface (e.g. `self_manage_list` paths reachable by an anonymous `Standard` session with ALLOW_ALL) without affecting authenticated sessions.

### 3. Per-DID / allowlist relay trust instead of open-relay
Provide a way to gate relay on the **inner forward's authcrypt sender** rather than a blanket `SEND_FORWARDED` in the global default:
- an allowlist of trusted peer-mediator DIDs permitted to relay, and/or
- per-DID ACL `SEND_FORWARDED` instead of via `global_acl_default`.

This lets an operator run a production relay between known mediators without an open-relay posture.

### 4. Revisit auto-registration of forwarding senders (fix #4 follow-up)
PR #383 now persists every forwarding sender as a `KNOWN_DID` seeded from `global_acl_default` (mirroring the recipient branch) so relay is reproducible. With an allowlist/per-DID model (#3) reconsider whether unknown forwarding senders should be auto-registered with ALLOW_ALL at all, or registered with a minimal/relay-only ACL. Watch for unbounded `KNOWN_DIDS` growth from arbitrary relay senders.

### 5. Documentation accuracy
- Update the `ACL_MODE:` doc line in `message_inbound.rs` — enforcement is per-message-type (to-mediator path) and per-envelope-sender (direct-delivery path), not a single session-level `SEND_MESSAGES` check.
- Document the relay trust model and the operator opt-in clearly (config docs + ADR if the trust model in #3 is adopted).

## Acceptance criteria
- [ ] Unauthenticated `/inbound` is accepted only for the relay forward path, with a minimal relay-scoped ACL.
- [ ] Invalid/expired/blocked tokens and backend errors are no longer silently downgraded to anonymous.
- [ ] Operators can run relay between an allowlist of trusted mediator DIDs without an open-relay posture.
- [ ] Shipped secure default still rejects unauthenticated `/inbound` (regression test retained).
- [ ] Unit/integration tests for the relay-scoped anonymous session and the allowlist gate; `cross_mediator_forwarding` example still round-trips reproducibly.

## Non-goals
- Re-architecting DIDComm forward/routing semantics.
- Changing the same-mediator (single-wrap) delivery path.

stormer78:
## Summary

Final piece of #385. Adds **per-hop re-wrapping** for inter-mediator relay, behind a new `relay_mode` config (default `blind` — **no behaviour change** for existing deployments).

> **Draft / verification needed:** the unit surface is fully covered and all feature combos build clean, but I could **not** run the live two-mediator + Redis `cross_mediator_forwarding` round-trip in my environment. This needs a manual end-to-end pass before merge — see *Manual verification* below.

## Background — why this exists

During #385 we established (by reading the SDK + mediator) that with the current **blind** relay:
- `forward_message` authcrypts the inner forward `from = the original sender`, and the relaying mediator forwards that inner envelope **byte-for-byte** (`ForwardQueueEntry { message: data.clone() }`).
- So the receiving mediator never sees the relaying *peer mediator's* identity — only the original sender's. That made item #3's "trusted peer-mediator allowlist" impossible at the DIDComm layer.
- Worse, an authcrypt JWE carries the sender's key id (`skid`) **in cleartext** in its protected header, so an observer on the mediator↔mediator hop can link sender → recipient-mediator.

Per maintainer decision, the fix is **per-hop encapsulation**: each mediator re-encrypts the inner envelope for the next hop, authcrypted *from itself*. This hides the original sender (and the inner envelope) from the wire **and** gives the receiving mediator an authenticated peer identity to allowlist.

## What's implemented

**Config (`[processors.forwarding]`)**
- `relay_mode = "blind" | "rewrap"` (default `blind`).
- `relay_trusted_mediators` — comma-separated allowlist of peer-mediator DIDs (rewrap only).

**Send side — `routing.rs::rewrap_for_relay`**
Wraps the inner forward attachment in a fresh `forward` authcrypted **FROM this mediator TO the next hop**, carrying the running `hop_count`, and enqueues *that* instead of the verbatim inner envelope.

**Receive side — `inbound.rs::peel_relay_rewrap_layers`**
A pre-pass (only in rewrap mode) that strips `forward`-to-self layers a peer produced: authenticates the relaying peer against the allowlist, then continues on the inner envelope. Bounded by `max_hops`. Implemented as a **pre-pass so `handle_inbound_didcomm`'s body is unchanged** — the direct-delivery and normal-message paths carry zero regression risk. (Cost: one extra unpack of the outer layer for to-mediator messages on a rewrap mediator — acceptable for an opt-in posture.)

**Trust gate — `relay_peer_trusted`**
Empty allowlist → accept any peer (ACL still applies). Non-empty → members only; an anonymous peer is rejected. This is the item-3 payoff, possible only because the rewrap layer's authcrypt `from` is the peer mediator.

## Trust ⟷ privacy note
Rewrap uses **authcrypt-from-mediator** (the chosen tradeoff): it reveals "which mediator relayed" to the receiving mediator in exchange for an authenticated, allowlistable peer identity. It hides the original sender + inner envelope from on-wire observers. (Pure anonymity — anoncrypt per hop — would forgo the peer identity and thus the allowlist; not chosen.)

## Tests
113 mediator lib tests pass (`--features redis-backend`), incl. new unit tests:
- `relay_mode` parsing (case-insensitive, unknown→blind, trimmed allowlist split)
- `relay_peer_trusted` (empty=any, membership-only, anon rejected)
- `rewrap_inner_attachment` (peels forward-to-self, ignores other-next & non-forward)

`clippy --all-targets -- -D warnings` clean on `redis-backend` and `didcomm,memory-backend`; `cargo deny` ok.

## Manual verification (before un-drafting / merge)
Two `did:peer` mediators on :7037/:7038 with `relay_mode = "rewrap"` (and optionally each other's DID in `relay_trusted_mediators`), then:
```
cargo run -p affinidi-messaging-helpers --example cross_mediator_forwarding -- \
  --alice-environment alice --bob-environment bob
```
Expected: both directions still round-trip; on the wire the mediator↔mediator envelope is authcrypted `from = relaying mediator` (no original-sender `skid`); with a populated allowlist, a non-listed peer's relay is rejected (`authorization.relay.untrusted_peer`).

## Commits
1. `feat: add relay_mode (blind|rewrap) config …`
2. `feat: implement rewrap relay — per-hop re-encryption + peer-mediator trust`

(Builds on #386 and #387, both merged.)

Closes #385

### DIFF
--- a/crates/messaging/affinidi-messaging-mediator/CHANGELOG.md
+++ b/crates/messaging/affinidi-messaging-mediator/CHANGELOG.md
@@ -4,6 +4,41 @@
 
 ## 10th June 2026
 
+### 0.15.20 — Per-hop re-wrapping for inter-mediator relay (#385 item 3, #388)
+
+- Final piece of #385. Adds **per-hop re-wrapping** for the inter-mediator
+  relay path, behind the new `[processors.forwarding] relay_mode` config
+  (`blind` | `rewrap`, default `blind` — **no behaviour change** for
+  existing deployments).
+- **Why:** in the historical *blind* relay the relaying mediator forwards
+  the inner envelope byte-for-byte, so (a) the receiving mediator never
+  sees the relaying *peer's* identity — making a trusted-peer allowlist
+  impossible at the DIDComm layer — and (b) the inner authcrypt JWE carries
+  the original sender's key id (`skid`) in cleartext on the
+  mediator↔mediator hop. In `rewrap` mode each mediator re-encrypts the
+  inner forward in a fresh `forward` authcrypted *from itself* to the next
+  hop: the original sender and inner envelope are hidden from on-wire
+  observers, and the receiver gets an authenticated peer identity to
+  allowlist.
+- **Send** (`routing.rs::rewrap_for_relay`): wraps the inner attachment in a
+  new `forward` authcrypted from this mediator to the next hop (carrying the
+  running `hop_count`), enqueued instead of the verbatim bytes.
+- **Receive** (`inbound.rs::peel_relay_rewrap_layers`): a pre-pass (rewrap
+  mode only) that strips `forward`-to-self layers a peer produced,
+  authenticates the relaying peer against `relay_trusted_mediators`
+  (empty = any; non-empty = members only, anonymous rejected with
+  `authorization.relay.untrusted_peer`), and continues on the inner
+  envelope. Bounded by `max_hops`. Implemented as a pre-pass so
+  `handle_inbound_didcomm`'s body is unchanged — zero regression risk on the
+  direct-delivery and normal-message paths; cost is one extra unpack of the
+  outer layer for to-mediator messages on a rewrap mediator.
+- Both mediators in a relaying pair must use the same mode. Verified
+  end-to-end on the memory backend by `affinidi-messaging-test-mediator`'s
+  `cross_mediator_forwarding` suite (rewrap round trip + trusted-peer
+  admit/reject), now possible because #399 made the forwarding processor
+  run on non-Redis backends; the on-wire crypto is covered by
+  `tests/relay_rewrap.rs`. Closes #385.
+
 ### 0.15.19 — Forwarding runs on Fjall and in-memory backends (#399)
 
 - The forwarding processor now spawns for every storage backend, not just
--- a/crates/messaging/affinidi-messaging-mediator/Cargo.toml
+++ b/crates/messaging/affinidi-messaging-mediator/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "affinidi-messaging-mediator"
-version = "0.15.19"
+version = "0.15.20"
 description = "Messaging Mediator service for Affinidi Messaging (DIDComm and TSP)"
 edition.workspace = true
 authors.workspace = true
--- a/crates/messaging/affinidi-messaging-mediator/affinidi-messaging-mediator-common/CHANGELOG.md
+++ b/crates/messaging/affinidi-messaging-mediator/affinidi-messaging-mediator-common/CHANGELOG.md
@@ -4,6 +4,15 @@
 
 ## 10th June 2026
 
+### 0.15.5 — `RelayMode` config for inter-mediator relay (#385 item 3)
+
+- Adds the `RelayMode` enum (`Blind` | `Rewrap`, default `Blind`) and two
+  fields on `ForwardingConfig`: `relay_mode` and `relay_trusted_mediators`
+  (a peer-mediator DID allowlist used only in `Rewrap`). Purely additive —
+  `Default` keeps the historical blind-relay behaviour, so existing
+  deployments are unchanged. Consumed by the mediator's routing/inbound
+  rewrap paths (see `affinidi-messaging-mediator` 0.15.20).
+
 ### 0.15.4 — `ForwardingProcessor` compiles for all storage backends
 
 - `tasks::forwarding::processor` is no longer gated on `redis-backend`.
--- a/crates/messaging/affinidi-messaging-mediator/affinidi-messaging-mediator-common/Cargo.toml
+++ b/crates/messaging/affinidi-messaging-mediator/affinidi-messaging-mediator-common/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "affinidi-messaging-mediator-common"
-version = "0.15.4"
+version = "0.15.5"
 description = "Shared types for the Affinidi Messaging Mediator (errors, database handler, config)"
 edition.workspace = true
 authors.workspace = true
--- a/crates/messaging/affinidi-messaging-mediator/affinidi-messaging-mediator-common/src/tasks/forwarding/config.rs
+++ b/crates/messaging/affinidi-messaging-mediator/affinidi-messaging-mediator-common/src/tasks/forwarding/config.rs
@@ -12,6 +12,30 @@
 use ahash::AHashSet as HashSet;
 use serde::{Deserialize, Serialize};
 
+/// How a mediator relays a forward whose next hop is a *remote* mediator.
+///
+/// See [`ForwardingConfig::relay_mode`]. Both mediators in a relaying pair must
+/// agree on the mode: a `Rewrap` sender produces envelopes only a `Rewrap`
+/// receiver knows how to peel.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum RelayMode {
+    /// Relay the inner forward attachment to the next mediator **unchanged**
+    /// (byte-for-byte). Simple, but the inner envelope — including the original
+    /// sender's key id in its JWE protected header — is visible on the wire,
+    /// and the receiving mediator cannot identify the relaying peer mediator.
+    /// This is the historical behaviour and the default.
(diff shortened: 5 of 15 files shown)


stormer78:
## Added: real-crypto round-trip verification (`tests/relay_rewrap.rs`)

Since a full two-mediator + Redis e2e wasn't runnable in my environment, I added an integration test that exercises the **re-wrap layer's cryptographic round-trip with real authcrypt** over the existing `did:peer` fixtures (ALICE = relaying mediator, BOB = receiving mediator). It builds the envelope exactly as `rewrap_for_relay` does, then unpacks it as the receiving mediator. Both tests pass.

It verifies the three properties the feature exists for:
1. **Privacy** — neither the inner envelope nor its base64 form appears in the outer JWE (the inner sender identity is encrypted, not on the wire).
2. **Peer identity** — the receiving mediator authenticates the **relaying mediator** as the authcrypt sender (`encrypted_from_kid` / `msg.from` == relaying mediator). This is precisely what blind relay can't do and what makes `relay_trusted_mediators` enforceable.
3. **Peel + loop continuity** — the `next == receiving-mediator` signal is present, `hop_count` survives the re-wrap, and the inner envelope is recovered byte-intact.

It also incidentally confirms `rewrap_for_relay`'s exact construction (no `created_time`/`expires_time`) packs and unpacks cleanly.

### Remaining gap
This covers the novel/risky **crypto + wire-format + peel** layer. The surrounding **HTTP + `FORWARD_Q` + forwarding-processor** plumbing is unchanged from the already-merged blind-relay path, so it's not re-exercised here. A full two-mediator + Redis `cross_mediator_forwarding` run (with `relay_mode = "rewrap"` on both) is still the recommended pre-merge smoke test. Leaving the PR as **draft** pending that, but the highest-risk part is now covered.

stormer78:
### Updated — rebased on `main` and the manual verification is now automated ✅

This branch was **rebased onto current `main`** (picking up #392–#400; the rewrap commits replayed with no conflicts) and the *Manual verification* step from the original description is now an **automated end-to-end suite** — so this is ready to un-draft.

**Why the live two-mediator run is finally possible without Redis:** the original blocker was that the forwarding processor + `forward_queue_*` were Redis-only, so a non-Redis two-mediator relay couldn't run. **#399** made the processor backend-agnostic (runs on memory/Fjall), and **#400** added an in-process two-mediator fixture. Rewrap rides on exactly that path, so it can now be exercised as a plain `#[tokio::test]`.

**New end-to-end coverage** (`affinidi-messaging-test-mediator`, memory backend, no Redis) — three cases added to the `cross_mediator_forwarding` suite, exercising the full rewrap plumbing (`routing::rewrap_for_relay` → `FORWARD_Q` → forwarding processor → loopback HTTP → `inbound::peel_relay_rewrap_layers`):

| Test | Asserts |
|---|---|
| `rewrap_relay_round_trips_end_to_end` | Both mediators in `Rewrap`, empty allowlist — Alice ⇄ Bob round-trips |
| `rewrap_relay_admits_trusted_peer` | B's `relay_trusted_mediators` names A → relay admitted, delivered |
| `rewrap_relay_rejects_untrusted_peer` | B's allowlist omits A → relay refused at B's peel with `authorization.relay.untrusted_peer` (HTTP 403, confirmed in logs); nothing reaches Bob |

The admit/reject pair is a controlled experiment (identical except the allowlist contents) isolating the allowlist as the gate. The same #400 double-forward construction drives blind and rewrap unchanged — only the mediator config differs. The on-wire crypto properties remain covered by the mediator crate's `tests/relay_rewrap.rs`.

**Fixture support added:** `TestMediatorBuilder::relay_mode(RelayMode)` + `relay_trusted_mediators(...)` (and a `RelayMode` re-export).

**Version bumps** (additive; `major.minor` pins unchanged):
- `affinidi-messaging-mediator-common` 0.15.4 → **0.15.5** (`RelayMode` + the two `ForwardingConfig` fields)
- `affinidi-messaging-mediator` 0.15.19 → **0.15.20** (routing rewrap + inbound peel pre-pass + config parse)
- `affinidi-messaging-test-mediator` 0.2.5 → **0.2.6** (builder knobs + e2e tests)

**Verification:** `cargo fmt --check`; `cargo clippy` clean (test-mediator `--all-targets`; mediator on `redis-backend` and `didcomm,memory-backend`); `cargo test -p affinidi-messaging-mediator --lib` 119 passed + `relay_rewrap` 2 passed; `-p affinidi-messaging-test-mediator` full suite incl. 5 `cross_mediator_forwarding` tests passed; `cargo deny` ok.

Issue #483, opened 2026-06-16
affinidi-tdk-common: `create_http_client` calls `Verifier::new_with_extra_roots`, which rustls-platform-verifier 0.7 doesn't provide on Android (breaks Android cross-compile)


stormer78:
## Summary

`affinidi-tdk-common::create_http_client` fails to compile for **Android** targets because it calls `rustls_platform_verifier::Verifier::new_with_extra_roots(..)`, an associated function that `rustls-platform-verifier 0.7.x` does **not** provide in its Android backend (only `new`). The call is not `cfg`-gated per platform, so any downstream crate that cross-compiles to an Android target fails to build.

Affects the latest published version, **`affinidi-tdk-common 0.6.4`** (and 0.6.3).

## Error

```
error[E0599]: no associated function or constant named `new_with_extra_roots` found for struct `Verifier` in the current scope
   --> .../affinidi-tdk-common-0.6.4/src/lib.rs:131:23
note: if you're trying to build a new `Verifier`, consider using `Verifier::new` which returns Result<Verifier, rustls::Error>
   --> .../rustls-platform-verifier-0.7.0/src/verification/android.rs:70:5
error: could not compile `affinidi-tdk-common` (lib) due to 1 previous error
```

## Root cause

`affinidi-tdk-common/src/lib.rs` (0.6.4):

- L72: `use rustls_platform_verifier::{ConfigVerifierExt, Verifier};`
- L131 (inside `create_http_client`, taken when `extra_roots` is non-empty):
  ```rust
  let verifier =
      Verifier::new_with_extra_roots(extra_roots.iter().cloned(), crypto_provider.clone())
          .map_err(|e| ...)?;
  ```

In `rustls-platform-verifier 0.7.0`, `new_with_extra_roots` is implemented for **some** platform backends but not all:

| backend (`src/verification/`) | `new` | `new_with_extra_roots` |
|---|:---:|:---:|
| `others.rs` (Linux/etc.) | ✅ | ✅ |
| `apple.rs` | ✅ | ✅ |
| `windows.rs` | ✅ | ✅ |
| **`android.rs`** | ✅ | ❌ |

So host builds (Linux/macOS/Windows) compile fine and the breakage is **Android-only** — which is why it slips past normal `cargo build`/CI and only surfaces when cross-compiling the engine for Android.

## Reproduction

Cross-compile any crate that depends on `affinidi-tdk-common` to an Android target, e.g.:

```
[...]

PR #486, opened 2026-06-17
fix: make did:cheqd opt-in so the resolver SDK no longer forces the rustls ring backend

stormer78:
## Problem

A downstream binary that links `affinidi-did-resolver-cache-sdk` panics at the
first TLS call:

```
thread 'main' panicked at 'no process-level CryptoProvider available'
```

This fires from `ClientConfig::builder()` (e.g. when `kube-client` initialises
its HTTP client) before any network connection is made, because `rustls 0.23`
has **both** the `aws_lc_rs` and `ring` backends compiled in and cannot
auto-select one.

## Diagnosis (verified in this workspace)

`cargo tree -i rustls@0.23 -e features` pinpoints the forcing edge:

```
rustls feature "ring"
  └── tokio-rustls feature "ring"
      └── tonic v0.12.3
```

`tonic 0.12.3` hardcodes `features = ["ring"]` on `tokio-rustls`, which forces
the `ring` feature onto `rustls 0.23` via feature unification. `tonic 0.12.3`
enters the build through **`did-resolver-cheqd v1.0.1`**, reached two ways:

1. directly — the SDK's `did-cheqd` feature (part of the default `did-methods`);
2. transitively — the SDK's `did-scid` feature → `did-scid`, whose own
   `default` enabled `did-cheqd` → `did-resolver-cheqd`.

When the consumer also brings `aws_lc_rs` (kube/reqwest/jsonwebtoken), both
backends are compiled → panic.

**Sibling sweep (task item #5):** the only other tonic/gRPC path is
mediator-common's opt-in `secrets-gcp` (`google-cloud-*`). Under that feature
tonic isn't even pulled and rustls stays `aws_lc_rs`-only — it does **not**
force ring. So the cheqd path is the sole offender.

## Why not just upgrade tonic?

`did-resolver-cheqd 1.0.1` is the **latest published version** and pins
`tonic 0.12`. Upgrading tonic requires an upstream change to an external crate,
so option (a) is not feasible from this workspace. `tonic 0.12` hardcodes
`ring`, so it can't be decoupled either (option b). The durable in-workspace
fix is option (c): stop forcing the backend by making the cheqd/tonic path
**opt-in**, so a single backend is compiled per build.

Per the workspace convention (fix the cause, not the symptom), **no runtime
`install_default()` was added** — installing a process-global rustls
`CryptoProvider` is the application's decision and stays in downstream `main()`.

## Changes

- **did-scid 0.1.10** — drop `did-cheqd` from `default` (now `["did-webvh"]`);
  gate the cheqd code paths and tests behind the feature so it compiles without
  it (returns a clear `CheqdError` for a `did:cheqd` source when disabled).
- **affinidi-did-resolver-cache-sdk 0.8.11** — drop `did-cheqd` from
  `did-methods` (now `["did-webvh", "did-scid"]`); consume `did-scid` with
  `default-features = false, features = ["did-webvh"]`. README documents the
  `did-cheqd` opt-in and the `ring` implications.

Both are **patch bumps**, keeping the `0.1`/`0.8` pins and `[patch.crates-io]`
redirects valid per ADR-0003.

## Proof

```text
# default + network (the panic-prone combo) — ring is GONE:
$ cargo tree -p affinidi-did-resolver-cache-sdk --features network -i rustls@0.23 -f "{f}"
aws-lc-rs,aws_lc_rs,log,logging,std,tls12

# whole workspace — no 'rustls feature ring' edge, tonic 0.12.3 absent entirely.

# opt-in still works (ring expected):
$ cargo tree -p affinidi-did-resolver-cache-sdk --features network,did-cheqd -i rustls@0.23 -f "{f}"
aws-lc-rs,aws_lc_rs,log,logging,ring,std,tls12
```

- `cargo check --workspace --all-targets` — green.
- `cargo clippy` on did-scid and the SDK, **both** with and without `did-cheqd` — green.
- did-scid lib tests pass in both feature configs (cheqd tests gated).

## Behaviour change

`did:cheqd` resolution is no longer available by default. Consumers that need it
add `features = ["did-cheqd"]` (which re-enables the `ring` backend) and install
a `CryptoProvider` in their binary's `main`.

### DIFF
--- a/crates/identity/affinidi-did-resolver-cache-sdk/CHANGELOG.md
+++ b/crates/identity/affinidi-did-resolver-cache-sdk/CHANGELOG.md
@@ -30,6 +30,12 @@
   cannot be fixed from this workspace (1.0.1 is the latest published version and
   `tonic 0.12` hardcodes `ring`); making cheqd opt-in is the durable in-workspace
   mitigation. Patch bump keeps the `0.8` pin valid.
+- **`rustls-platform-verifier` bumped `0.6` → `0.7`** to match the rest of the
+  workspace (`affinidi-tdk-common` is already on `0.7`), consolidating the lock
+  to a single version. The SDK only calls `ClientConfig::with_platform_verifier()`
+  (available on all backends, incl. Android) and does not use
+  `Verifier::new_with_extra_roots`, so the Android cross-compile gap from #483/#484
+  does not apply here. No source change required.
 
 ## 14th June 2026
 
--- a/crates/identity/affinidi-did-resolver-cache-sdk/CHANGELOG.md
+++ b/crates/identity/affinidi-did-resolver-cache-sdk/CHANGELOG.md
@@ -2,6 +2,35 @@
 
 ## Changelog history
 
+## 17th June 2026
+
+### 0.8.11 — `did-cheqd` is now opt-in (no forced rustls `ring` backend)
+
+- **`did-cheqd` removed from the default `did-methods` set.** It pulled
+  `did-resolver-cheqd`, whose `tonic 0.12` dependency hardcodes the rustls
+  `ring` backend on `tokio-rustls`/`rustls 0.23`. Combined with `network`
+  (which uses `aws_lc_rs`) — or any downstream binary that selects `aws_lc_rs`
+  via `kube`/`reqwest`/`jsonwebtoken` — both rustls backends were compiled and
+  `rustls` could no longer auto-select one, panicking with "no process-level
+  CryptoProvider available" at the first TLS call (e.g. `ClientConfig::builder()`).
+  `did-methods` is now `["did-webvh", "did-scid"]`; a default + `network` build
+  compiles `aws_lc_rs` only.
+- **The `did-scid` dependency now uses `default-features = false` + `did-webvh`**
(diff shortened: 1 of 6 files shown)


stormer78:
Added a follow-up commit: **bump `rustls-platform-verifier` 0.6 → 0.7** in `affinidi-did-resolver-cache-sdk`.

- Aligns the SDK with the rest of the workspace (`affinidi-tdk-common` is already on 0.7), **consolidating the lock from two versions (0.6.2 + 0.7.0) down to a single 0.7.0**.
- **No source change required** — the SDK only calls `ClientConfig::with_platform_verifier()` (signature unchanged, available on all backends incl. Android) and does *not* use `Verifier::new_with_extra_roots`, so the Android cross-compile gap from #483/#484 does not apply here.
- Verified: SDK clippy clean, full `cargo check --workspace --all-targets` green, rustls still `aws_lc_rs`-only (no `ring`) on the default+network build, and `cargo tree -i rustls-platform-verifier` now shows a single 0.7.0.

Other external deps in both crates already resolve to their latest compatible releases within their caret ranges; `did-resolver-cheqd` is already at its latest (1.0.1), which is exactly why the cheqd path had to be made opt-in rather than upgraded.

PR #494, opened 2026-06-22
chore: ignore .env files

geoffturk:
Add .env and .env.* to .gitignore so operator-local secrets and credentials from downstream build pipelines are never accidentally checked in.


stormer78 (approved):
LGTM

PR #526, opened 2026-06-24
fix: serve did:web-native document at /.well-known/did.json for self-hosted did:webvh mediators

marat-affinidi:
## Problem

A mediator configured with `did_web_self_hosted = file://.../did.jsonl` (the default when set up via `mediator-setup --did-method webvh --save-did-web`) serves the DID Document extracted from its **did:webvh** log at `/.well-known/did.json`. That document's `id` — and every verification-method and service self-reference — stays under `did:webvh:{scid}:{domain}`.

`/.well-known/did.json` is the **did:web** resolution surface (did:webvh resolvers fetch `/.well-known/did.jsonl`, per spec). So a resolver looking up `did:web:{domain}` receives a document whose `id` does **not** match the DID it resolved, and whose keys are published as `did:webvh:{scid}:{domain}#key-0` — meaning `did:web:{domain}#key-0` cannot be dereferenced. The Affinidi did:web resolver does not reject the mismatched `id`, but downstream key/service dereferencing breaks.

Observed live:
```
curl -sf https://<mediator>/.well-known/did.json | jq .id
# -> "did:webvh:Qm...:<domain>"   (should be "did:web:<domain>")
```

## Fix

When the self-hosted source is a did:webvh log, rewrite the **`did.json`** body's identifier `did:webvh:{scid}:{domain}` -> `did:web:{domain}` (id, controller, key ids, service ids). Left verbatim:
- `/.well-known/did.jsonl` — the did:webvh log stream
- the internal typed `Document` (`mediator_did_document`) — the mediator's real `did:webvh` identity, used for resolver preload

| Surface | `id` | Consumer |
|---------|------|----------|
| `/.well-known/did.json`  | `did:web:{domain}` (fixed) | did:web resolvers |
| `/.well-known/did.jsonl` | `did:webvh:{scid}:{domain}` | did:webvh resolvers |
| internal `Document` | `did:webvh:{scid}:{domain}` | mediator self / resolver cache |

### Implementation
- Hoist the canonical rewrite (`webvh_did_to_web`, `webvh_log_to_did_web`) out of the `mediator-setup` binary into `affinidi-messaging-mediator-common::did_web` (pure, `serde_json`-only, always-built) so the runtime and the wizard share **one tested implementation**. `mediator-setup` now re-exports it.
- Add `rewrite_did_document_to_web(state, webvh_did)` for the runtime, which already holds the extracted `state`.
- `config/mod.rs`: extract the self-hosted load path into a pure, unit-tested `split_self_hosted_did_source(...)` and build `mediator_did_doc` from the rewritten document.
- Doc comment on `well_known_did_json_handler` updated.

Automatic — no new config key, no operator action — and non-breaking for did:webvh resolvers.

## Testing

- **Unit** (`mediator-common::did_web`, 6): scid drop, path preservation, malformed rejection, full-document rewrite (id/controller/VM/service), log-envelope extraction, missing-`state` error.
- **Config** (`affinidi-messaging-mediator`, 2): feeds a real `didwebvh-rs` log vector through `split_self_hosted_did_source` and asserts `did.json` `id == did:web:example.com` (no `did:webvh:` traces), `did.jsonl` unchanged, typed `Document` keeps `did:webvh`; plus the plain did:web passthrough case.
- **mediator-setup**: existing did:web-export tests still pass.
- **End-to-end**: built the mediator from this branch, ran a self-hosted did:webvh mediator (`--save-did-web`), and confirmed:
  ```
  /.well-known/did.json  -> id       = did:web:mediator              (all VM/service ids did:web:)
  /.well-known/did.jsonl -> state.id = did:webvh:QmZde...:mediator    (verbatim)
  ```

`cargo build -p affinidi-messaging-mediator` + clippy clean.

### DIFF
--- a/crates/messaging/affinidi-messaging-mediator/CHANGELOG.md
+++ b/crates/messaging/affinidi-messaging-mediator/CHANGELOG.md
@@ -4,6 +4,19 @@
 
 ## 24th June 2026
 
+### 0.16.27 — did:web self-hosting: serve a valid did:web document
+
+- A self-hosted `did:webvh` mediator (`did_web_self_hosted = file://.../did.jsonl`)
+  now serves a **did:web-native** document at `/.well-known/did.json`: its identifier
+  (and every verification-method and service self-reference) is rewritten from
+  `did:webvh:{scid}:{domain}` to `did:web:{domain}`. Previously the extracted log
+  `state` was served verbatim, so a `did:web:{domain}` resolver received a document
+  whose `id` didn't match the DID it resolved and whose `#key-…` references couldn't
+  be dereferenced. The `/.well-known/did.jsonl` log stream and the mediator's internal
+  `did:webvh` identity are unchanged. Automatic — no new config key.
+
+## 24th June 2026
+
 ### 0.16.26 — TSP: Control message relay
 
 - The TSP inbound handler now relays **`Control`** messages (relationship invite /
--- a/crates/messaging/affinidi-messaging-mediator/Cargo.toml
+++ b/crates/messaging/affinidi-messaging-mediator/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "affinidi-messaging-mediator"
-version = "0.16.26"
+version = "0.16.27"
 description = "Messaging Mediator service for Affinidi Messaging (DIDComm and TSP)"
 edition.workspace = true
 authors.workspace = true
--- a/crates/messaging/affinidi-messaging-mediator/affinidi-messaging-mediator-common/CHANGELOG.md
+++ b/crates/messaging/affinidi-messaging-mediator/affinidi-messaging-mediator-common/CHANGELOG.md
@@ -2,6 +2,17 @@
 
 ## Changelog history
 
+## 24th June 2026
+
+### 0.15.18 — did:webvh → did:web document rewriting
+
+- New always-built `did_web` module: `webvh_did_to_web`, `rewrite_did_document_to_web`,
+  and `webvh_log_to_did_web` convert a `did:webvh:{scid}:{domain}` DID document into its
+  wire-compatible `did:web:{domain}` form (id, controller, key and service
+  self-references rewritten). Pure (`serde_json` only), shared by the mediator runtime
+  (serving `/.well-known/did.json`) and the `mediator-setup` wizard (writing the
+  `did-web.json` operator artefact) so both use one tested implementation.
+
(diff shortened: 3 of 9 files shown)


stormer78:
## Follow-up: hardened the rewrite + reconciled with `main`

Pushed two commits addressing review feedback and bringing the branch up to date with `main`.

### Rewrite hardening (`8f097ae`)
Replaced the blind whole-string `.replace(webvh_did, web_did)` in `rewrite_did_document_to_web` with a **structured recursive walk** that only rewrites genuine DID-URL self-references — the bare DID, or the DID followed by a `#fragment` / `?query` / `/path` component. This addresses three edge cases the string replace got wrong:

- **Foreign references survived** — a delegated `controller` (or any reference) pointing at a *different* `did:webvh:{otherScid}:{otherDomain}` was left as a `did:webvh:` trace. It's now left **verbatim on purpose**: rewriting another entity's DID to a `did:web` location we don't host would be incorrect. For the normal self-hosted case (controller = self) the served document is fully `did:web` as intended.
- **Over-matching corrupted embedded values** — a `serviceEndpoint` URL embedding the DID in a query string, or a longer prefix-sharing DID like `{did}:tenant`, were partially/wrongly rewritten. These are now untouched.
- **Empty SCID accepted** — `webvh_did_to_web` now rejects `did:webvh::domain` with a new `EmptyScid` error instead of silently producing a `did:web` DID.

New unit test `rewrite_leaves_foreign_and_substring_references_verbatim`; the existing whole-document and config tests still pass.

### Reconciled with `main` / #527 (`4007ef8`)
`main` advanced to #527 (TSPTransport service advertisement), which touched the same `config/mod.rs` area and **also bumped the mediator to `0.16.27`**. Merged and resolved:

- Kept both PRs' functions and test modules in `config/mod.rs`.
- **Version bump:** resolved the `0.16.27` collision by bumping this PR's mediator entry to **`0.16.28`** and reorganising the CHANGELOG. `mediator-common` stays at `0.15.18` (no collision); no other crate pins the mediator binary version, so no cascade.
- Collapsed a let-chain in #527's merged code to clear a newer-clippy `collapsible_if` warning under `-D warnings`.

### Verification
- `cargo build` + clippy clean on **both** feature sets (default and `tsp`).
- Tests green: 7 `did_web` unit, 37 `tsp` config (incl. #527's), 360 `mediator-setup` (re-export intact).

Branch is now current with `main` and conflict-free.

PR #541, opened 2026-06-28
fix(mediator): treat empty sender hash as anonymous in store_message

maratsh:
An anonymous routing/2.0/forward (no `from`) reaches store_message with an empty sender hash instead of the ANONYMOUS sentinel. The shared from_hash.unwrap_or("ANONYMOUS") normalization only catches None, so the empty string builds an empty store key: the fjall backend asserts non-empty keys and panics; redis/memory silently store a malformed empty-keyed entry. Guard all three backends (fjall, memory, redis) to treat an empty hash as anonymous, plus a fjall regression test. Unblocks block_anonymous_outer_envelope=false (spec-standard anonymous forwards).


stormer78 (approved):
LGTM

PR #547, opened 2026-06-29
fix: recursive infinite loop

YoussefAWasfy:


stormer78:
## Updated: synced to `main`, fixed the build break, added a regression test

Rebuilt this branch on top of the latest `main` as two clean linear commits. Summary of what changed and why, for anyone picking this up:

### The core fix (kept as-is)
`RedisStore`'s `MediatorStore::oob_discovery_get` / `oob_discovery_delete` called `self.oob_discovery_get(..)` / `self.oob_discovery_delete(..)` — which resolve back to *themselves* (there's no inherent method of that name), so every OOB-invite **claim/delete recursed infinitely → stack overflow**. They now issue the underlying Redis ops directly (`HMGET` / `DEL`), mirroring `oob_discovery_store`, with a best-effort `OOB_INVITES_CLAIMED` bump that never fails the request.

### What I removed
The branch had also bundled a **half-finished, unrelated refactor of `affinidi-tsp::decode_payload_frame`** — it deleted the `DecodedFrame` struct + `hops_to_strings` helper and changed the return type to a tuple, but never updated the function body or callers (still building `DecodedFrame { .. }` at 5 sites). That's what was failing `cargo check` / clippy / tests / coverage / verify-publish (`error[E0422]: cannot find struct ... DecodedFrame`). It's unrelated to the recursion fix, so I reverted `direct.rs` to `main`'s working version.

### New: regression test (the important bit)
The recursion slipped through because **`tests/store_conformance.rs` — the backend-agnostic suite that runs against memory, fjall *and* Redis — had no `oob_discovery` coverage at all**. The bug is Redis-only (memory/fjall have separate, correct impls), and the compiler doesn't flag trait-method self-recursion, so nothing caught it.

Added an **`oob_discovery_roundtrip`** check (store → get → delete → get-is-gone → delete-of-missing), registered against all three backends (Redis on DB index `@ 12`). Verified against a live Redis:
- original recursive impl → `thread 'redis::oob_discovery_roundtrip' has overflowed its stack / fatal runtime error: stack overflow, aborting` (SIGABRT) ✗
- fixed impl → passes ✓

### Heads-up: `main` was red on fmt
Independently, `main` itself is currently failing `cargo fmt --all --check` at 10 locations (`affinidi-tsp` `lib.rs`/`message/envelope.rs`/`message/wire.rs`/`examples/alice_and_bob.rs` and the two mediator stores). Any branch inherits that red check, so this PR includes the `cargo fmt` cleanup to go green — pure reformatting, no behaviour change. (This is why those files show up in the diff.)

### Version bumps
- `affinidi-messaging-mediator-common` 0.15.20 → **0.15.21** (the fix)
- `affinidi-messaging-mediator` 0.16.34 → **0.16.35** (conformance test + store fmt)
- `affinidi-tsp` 0.1.11 → **0.1.12** (fmt only)

Dropped the unrelated `trust-tasks-rs` `"0.2"` → `"0.2.12"` pin tightening (no-op — `"0.2"` already resolves there).

### Verified locally
`cargo fmt --all --check`, `clippy --all-features --tests -D warnings`, the version-bump release guard, and the full `store_conformance` suite (memory + fjall + live Redis) all pass.

> Note on the OOB expiry in the test: it uses a far-future absolute timestamp (`4_102_444_800`, year 2100) rather than `0`, because Redis stamps the key with `EXPIREAT` — `0`/past would expire it immediately, while the in-process backends treat `0` as "never expires". Far-future keeps it live on all three.

PR #586, opened 2026-07-03
fix: negotiate shared key-agreement curve when packing encrypted mediator replies

marat-affinidi:
After migrating clients to the P-256 key suite, the mediator fails to pack encrypted (authcrypt/anoncrypt) replies to P-256 recipients:

```
e.p.message.pack / HTTP 400
"key agreement failed: curve mismatch between private and public keys"
```

`pack_encrypted` selected the recipient's first key-agreement key (X25519) regardless of the sender's curve. This negotiates a shared curve for authcrypt and picks a curve-matched recipient key for anoncrypt, with regression tests covering both paths.

### DIFF
--- a/crates/messaging/affinidi-messaging-mediator/src/didcomm_compat.rs
+++ b/crates/messaging/affinidi-messaging-mediator/src/didcomm_compat.rs
@@ -14,6 +14,7 @@
 use affinidi_crypto::jose::key_agreement::{Curve, PrivateKeyAgreement, PublicKeyAgreement};
 use affinidi_did_common::{
     document::DocumentExt,
+    key_negotiation::{DEFAULT_CURVE_PREFERENCE, negotiate_authcrypt, select_anoncrypt_key},
     verification_method::{VerificationMethod, VerificationRelationship},
 };
 use affinidi_did_resolver_cache_sdk::DIDCacheClient;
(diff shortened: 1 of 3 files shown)


crates/messaging/affinidi-messaging-mediator/src/didcomm_compat.rs
```diff
@@ -676,66 +677,97 @@ fn resolve_public_key(
...
+            let Ok(curve) = key_type_to_curve(secret.get_key_type()) else {
+                continue;
+            };
+            match PrivateKeyAgreement::from_raw_bytes(curve, secret.get_private_bytes()) {
+                Ok(private) => sender_keys.push((kid, private, curve)),
+                Err(_) => continue,
+            }
+        }
+        if sender_keys.is_empty() {
+            return Err("Sender has no usable key agreement key".to_string());
```
anton-i-affinidi: It might be useful to show what is "usable" and what sender has.

marat-affinidi: Good call — done in 2d192aa. The error now lists what the sender advertised and why each key was unusable, e.g.:

```
Sender has no usable key-agreement key (a usable key needs a held secret on a supported curve). Advertised: [did:example:sender#key-1]; unusable: [did:example:sender#key-1 (no secret held)]
```

Per-key reasons are `no secret held` / `unsupported key type: <T>` / `invalid key material: <e>`. Added `pack_encrypted_authcrypt_reports_unusable_sender_keys` to lock it in.

anton-i-affinidi (approved):

stormer78 (approved):
LGTM

PR #589, opened 2026-07-08
feat: add read-only secret-backend probe

marat-affinidi:
## Summary

Adds `SecretStore::probe_readonly()` to `affinidi-messaging-mediator-common`: a default-implemented reachability + credential check that reads a fixed sentinel key and **never mutates** the backend.

Unlike the existing write→read→delete `probe()`, a caller with read-only credentials can use it. An absent sentinel (`Ok(None)`) is the healthy case; only a transport/permission failure returns `Err`. On AWS this needs only `GetSecretValue` (`ResourceNotFound` → `Ok(None)`; auth failure → `Err`).

`MediatorSecrets` gains a matching `probe_readonly()` passthrough.

This is the upstream half of the change. Repointing the runtime probe call sites to `probe_readonly()` follows in a separate PR once this version is published, so the downstream crate resolves the new API from the registry.

## Testing
- `cargo publish --dry-run` clean for `affinidi-messaging-mediator-common`.
- `cargo clippy --all-targets` clean with `-D warnings --cfg tracing_unstable`.
- `cargo fmt --check` clean.
- `cargo test secrets::` — 77 passed (incl. 2 new `probe_readonly` tests).


crates/messaging/affinidi-messaging-mediator/affinidi-messaging-mediator-common/src/secrets/store.rs
```diff
@@ -101,6 +101,27 @@ pub trait SecretStore: Send + Sync {
             reason: format!("sentinel read OK but delete failed: {e}"),
         })
     }
+
+    /// Read-only reachability + credential check: proves the backend is
+    /// reachable and the caller is authenticated **without mutating it**.
```
stormer78, line 106: **Blocking.** This claim doesn't hold for three of the eight backends.

`probe_readonly()` cannot return `Err` for `memory`, `file`, or `file_encrypted`:

- `file::load_locked` maps `ErrorKind::NotFound` → `Ok(StoreFile::default())`, so a **deleted or mistyped store file reads as healthy**.
- `file_encrypted::get` reads a `Mutex<EncryptedStoreFile>` populated once at `open()` — it never touches the disk again.
- `memory::get` is infallible apart from mutex poisoning.

Today's `probe()` writes through `save_locked` → `create_dir_all` + `fs::write`, so it catches a read-only filesystem, a permission error, or a full disk. #591 repoints both `/readyz` and the boot check to this method, so none of those are caught any more for file-backed deployments, and `secrets_backend_reachable` becomes a constant `true`.

Suggest overriding `probe_readonly()` in `file` and `file_encrypted` so it actually touches the path — `fs::metadata(&self.path)`, or a full `load_locked()` for `file`. The default impl stays correct for the cloud and k8s backends, where `get` genuinely round-trips.

crates/messaging/affinidi-messaging-mediator/affinidi-messaging-mediator-common/src/secrets/store.rs
```diff
@@ -101,6 +101,27 @@ pub trait SecretStore: Send + Sync {
...
+    ///
+    /// Unlike [`SecretStore::probe`] (write → read → delete, requires write
+    /// permissions), this only reads a fixed sentinel key, so a read-only
+    /// caller can use it. An absent sentinel (`Ok(None)`) is healthy — it
+    /// still proves the backend answered with valid credentials; only a
+    /// transport/permission failure returns `Err`. On AWS this needs just
+    /// `GetSecretValue` (`ResourceNotFound` → `Ok(None)`; auth failure → `Err`).
```
stormer78: Worth stating the flip side here: because the sentinel is *never created*, this permanently reads a name that doesn't exist.

That's fine under a prefix-globbed grant (`Resource: arn:...:secret:{prefix}*`), which is what the PR description assumes. But a read-only role scoped to an **enumerated list of exact secret ARNs** — a common hardening pattern — gets `AccessDeniedException`, which `aws.rs` maps to `Unreachable`. So a correctly-provisioned, fully-functional role reports the backend as down. Same shape on GCP (per-secret IAM bindings) and Vault (per-path policies).

`probe()` has the same property, so it isn't a regression — but this is the method being sold as "works with read-only credentials", so the grant requirement (the prefix, not just the well-known keys) belongs in this doc and in `docs/secrets-backend.md`. Also worth telling operators that every `/readyz` poll will now emit a `ResourceNotFound`-shaped entry in CloudTrail / Cloud Audit Logs.

crates/messaging/affinidi-messaging-mediator/affinidi-messaging-mediator-common/src/secrets/store.rs
```diff
@@ -101,6 +101,27 @@ pub trait SecretStore: Send + Sync {
...
+    async fn probe_readonly(&self) -> Result<()> {
+        // Fixed name (not a UUID): only ever read, never created.
+        let key = format!("{PROBE_SENTINEL_PREFIX}readonly");
```
stormer78: Since the key is fixed rather than UUID-derived, this wants to be a `pub const` next to `PROBE_SENTINEL_PREFIX` in `well_known.rs`. It's duplicated verbatim in the test at L170, and it allocates on every probe — i.e. on every `/readyz` poll once #591 lands.

`concat!` won't take a `const`, so the cheapest form is a plain `pub const PROBE_READONLY_KEY: &str = "mediator_probe_readonly";` plus a test asserting `PROBE_READONLY_KEY.starts_with(PROBE_SENTINEL_PREFIX)`.

crates/messaging/affinidi-messaging-mediator/affinidi-messaging-mediator-common/src/secrets/store.rs
```diff
@@ -101,6 +101,27 @@ pub trait SecretStore: Send + Sync {
...
+        self.get(&key)
+            .await
+            .map(|_| ())
+            .map_err(|e| SecretStoreError::ProbeFailed {
```
stormer78: This diverges from `probe()`'s error taxonomy for the same underlying condition.

`probe()` propagates the backend's error verbatim on the read path (`self.get(&key).await?` → caller sees `SecretStoreError::Unreachable`) and only wraps into `ProbeFailed` for the delete step. Here *everything* is wrapped, so an unreachable Vault surfaces as `Unreachable` through one method and `ProbeFailed { reason: "read-only reachability probe failed: ..." }` through the other.

`/readyz` only logs `Display`, so the impact is low today, but a caller matching on the variant would be surprised. I'd propagate with `?` and drop the wrapper — the inner errors already carry `backend`.

Side effect of the current form: a genuinely malformed sentinel (`InvalidShape`, if someone ever *did* create that key with non-base64 content) reports as a probe failure rather than a data problem.

crates/messaging/affinidi-messaging-mediator/affinidi-messaging-mediator-common/src/secrets/store.rs
```diff
@@ -132,4 +153,21 @@ mod tests {
         let store: DynSecretStore = Arc::new(MemoryStore::new("memory"));
         store.probe().await.unwrap();
     }
+
+    #[tokio::test]
+    async fn probe_readonly_succeeds_on_empty_store() {
```
stormer78, line 172: Both new tests run against `MemoryStore`, whose `get` cannot fail. They confirm the default impl compiles and doesn't call `put` — worth something — but neither exercises a branch that can realistically break.

The untested path is the one that matters: a small stub store whose `get` returns `Err(Unreachable { .. })`, asserting the failure surfaces (and, if the `ProbeFailed` wrapper stays, that it carries the right `backend` and a `reason` containing the source).

The `Ok(Some(_))` branch is unexercised too. It's dead in practice since nothing ever writes the key, but `.map(|_| ())` says it's meant to pass.

Minor: `probe_readonly_does_not_mutate_the_store` only checks the sentinel key specifically. Asserting the store is still *empty* would be the stronger claim.

stormer78 (commented):
Reviewed alongside #591, which is what surfaced the main issue: the two runtime call sites it repoints (`Config::TryFrom` and `/readyz`) sit behind backends where `probe_readonly()` cannot fail.

The design is right — the trait method is in the correct place, and splitting the read-only probe out of the write-based one is well motivated. `/readyz` currently calls `probe()`, which means three mutating API calls per readiness poll and a runtime role that carries write permission purely for health checking.

**One blocking item**: `probe_readonly()` is a no-op for the `file`, `file_encrypted`, and `memory` backends, so the health signal it feeds becomes vacuously true. Detail inline on `store.rs`. Everything else below is polish that's cheap now, while the API is unreleased and has no callers.

### Things I checked that are right

- **No retry amplification on the healthy path.** The healthy case is `Ok(None)`, so every cloud backend takes its not-found branch. All four policies classify that as terminal — `AwsRetryPolicy` (documented short-circuit on `ResourceNotFoundException`), `GcpRetryPolicy` (`Code::NotFound => Retryable::No`), `AzureRetryPolicy` (`_ => No` under `HttpResponse`), `VaultRetryPolicy` (only 429/5xx retry) — so a missing sentinel returns immediately rather than burning 100 ms + 400 ms of backoff. Easy thing to get wrong.
- **The sentinel key is legal on every backend.** `mediator_probe_readonly` clears `validate_key` on GCP (no `/`) and Azure (no `/` or `.`), and Azure's `secret_name` maps `_` → `-`, so it lands as a valid Key Vault name. AWS and Vault accept underscores.
- **k8s is well served.** `k8s::get` calls `api.get_opt()` against the API server, so the probe is genuinely meaningful for the read-only-service-account deployment this work targets.

### Nit

The changelog jumps `0.15.21` → `0.15.26`; `.22`–`.25` shipped unlogged. Pre-existing, not this PR's doing, but the gap is visible in the diff.

crates/messaging/affinidi-messaging-mediator/affinidi-messaging-mediator-common/src/secrets/store.rs
stormer78: Following up with the concrete fix — and with a **correction to my comment above**, which over-claimed. Having traced each backend's `open()` path, the accurate picture is narrower than "three of the eight backends":

- **`memory` isn't operator-reachable.** `open_store` has no arm for it (`BackendUrl` has no `Memory` variant), so it's a library/test construct, not a deployment. Ignore it.
- **`file`'s *boot* check does hit the disk.** `FileStore::open` sets `state: Mutex::new(None)`, so the first `get` lazily calls `load_locked()` → `fs::read`. A corrupt store file (parse error) or an unreadable one (`PermissionDenied`) still fails at boot. What boot loses is only the *write-side* coverage `probe()` had: read-only filesystem, disk full, no write permission.
- **`file_encrypted`'s boot is fine regardless of the probe.** `open()` reads, parses, checks the version, derives the key, and runs a verifier check before returning.

So the boot regression is real but modest. **The part that genuinely breaks is the `/readyz` re-probe**, and that's the one `handlers/mod.rs` explicitly documents as existing to catch "mid-flight credential expiries / network blips that the boot-time probe couldn't":

- `file::get` is served from the in-memory `state` after the first load — it never re-reads the disk.
- `file_encrypted::get` reads a `Mutex<EncryptedStoreFile>` populated once at `open()` — it never reads the disk at all.

So once #591 repoints `/readyz`, `secrets_backend_reachable` is a constant `true` for both file backends for the life of the process, no matter what happens to the store on disk. That's still worth blocking on, since `/readyz` gates load-balancer membership.

---

### The fix

Two overrides. Both restore a real disk touch on every probe; neither disturbs the cache.

**`backends/file.rs`** — `load_locked()` already takes `&self` and re-reads from disk, so this is a one-liner:

```rust
async fn probe_readonly(&self) -> Result<()> {
    // `get` is served from `state` once the first load populates it, so the
    // trait default stops touching the disk after boot — `/readyz` could
    // never observe a store that vanished or became unreadable mid-flight.
    // Re-read from disk; deliberately don't refresh the cache, so the probe
    // stays free of side effects.
    self.load_locked().map(|_| ())
}
```

**`backends/file_encrypted.rs`** — here `get` never touches the disk, so the default is a pure in-memory no-op:

```rust
async fn probe_readonly(&self) -> Result<()> {
    // `state` is populated once by `open()` and `get` never re-reads, so the
    // trait default can't observe a vanished, truncated, or corrupted store.
    // Re-read and parse the envelope on every probe.
    //
    // Deliberately no KDF re-derivation: `open()` already verified the
    // passphrase, and running Argon2 on every `/readyz` poll would be a
    // self-inflicted DoS.
    let bytes = fs::read(&self.path).map_err(|source| SecretStoreError::Io {
        backend: BACKEND_LABEL,
        source,
    })?;
    serde_json::from_slice::<EncryptedStoreFile>(&bytes).map_err(|e| SecretStoreError::Io {
        backend: BACKEND_LABEL,
        source: std::io::Error::new(
            std::io::ErrorKind::InvalidData,
            format!("failed to parse encrypted store {}: {e}", self.path.display()),
        ),
    })?;
    Ok(())
}
```

That Argon2 point is the trap worth calling out explicitly: the "obvious" implementation of a read-only probe for an encrypted store is to re-run `open()`, which would put a memory-hard KDF on the readiness path.

### The one semantic decision

The two overrides treat a **missing file** differently, and I think that asymmetry is correct rather than an oversight — but it should be a deliberate call, and documented:

- `file` → **healthy** (`load_locked` maps `NotFound` → `Ok(StoreFile::default())`). `FileStore::open` doesn't create the file, and `mediator-setup` calls `open_and_probe_secret_backend` *before* provisioning, so first-run must not fail. This also mirrors the trait contract, where an absent cloud sentinel is the healthy case.
- `file_encrypted` → **error** (`fs::read` returns `NotFound`). `open()` calls `init_new` when the file is absent, so by the time any probe runs the file must exist. Its disappearance is a genuine fault.

Worth noting the old `probe()` never caught a missing `file://` store either — `put` → `save_locked` → `create_dir_all` + `fs::write` silently *created* it, arguably a bug of its own (a typo'd path yields a fresh empty store rather than an error). So "missing file reads as healthy" is not a regression introduced here; just don't let the `probe_readonly` doc claim otherwise.

### Tests

The gap worth closing is a probe that actually fails. For `file`, write `b"{ not json"` to a temp path, open the store, and assert `probe_readonly()` is `Err` — that pins the disk-touch and would fail today against the trait default (which returns `Ok(())` because nothing ever reads). For `file_encrypted`, open a store, delete the file underneath it, and assert `probe_readonly()` is `Err` while `get()` still cheerfully returns `Ok(None)` — that's the exact `/readyz` blind spot, made executable.

### No override needed

- **`keyring`** — `entry.get_password()` hits the OS secret service on every call; `NoEntry` → `Ok(None)`, a locked or absent service → `Err`. Correct as-is.
- **cloud + `k8s`** — `get` is a real network round-trip. The default impl is exactly right, which is the whole point of putting it on the trait.

### Cost

For file backends this makes `/readyz` do one `fs::read` + JSON parse per poll. That's strictly less work than today's `probe()`, which did a read *plus* two writes and a delete — so this is a net improvement on the hot path, not a new cost.

stormer78:
Pushed `dee8bc6` onto this branch — one fix found while reviewing.

**`probe_readonly` read the store file without taking the state lock.** `put`/`delete` hold that lock across `save_locked`, which does `OpenOptions::truncate(true).open()` and *then* `write_all`. An unlocked probe can land in that window:

- `file_encrypted`: reads zero bytes, `serde_json::from_slice::<EncryptedStoreFile>(&[])` fails, and `/readyz` reports `secrets_backend_reachable: false`.
- `file` (plaintext): benign — `load_locked` maps empty bytes to `Ok(default)` — but it is still judging the backend on a half-written store.

This is reachable in production rather than theoretical: in VTA mode the runtime writes the cached bundle at `common/config/vta_cache.rs:68`, driven both at boot and by the background refresh task (`tasks/vta_refresh.rs:116`) while the server is serving `/readyz`. Before #591 repoints it, `/readyz` called `probe()`, which went through `get`/`put`/`delete` and therefore serialised on the lock — so the pair of PRs is what opens the window.

The fix takes the same lock in both overrides. Each backend gets a test that holds the lock and asserts the probe blocks rather than reading through; both fail without the change (I removed the guards and re-ran to confirm they are load-bearing — a timing-based stress test I tried first passed either way, so it is not that).

`cargo test -p affinidi-messaging-mediator-common` → 175 passed. `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` clean with `--cfg tracing_unstable`, both default and `--all-features`.

No version bump: 0.15.26 is unpublished, so this folds into it. The `0.15.21` → `0.15.26` changelog gap is still there (pre-existing, `.22`–`.25` shipped unlogged) if you want to backfill it while this is open.

stormer78 (approved):
Approving. The design is right — putting the read-only probe on the trait, defaulting it to a `get` of a never-written sentinel, and letting the backends whose `get` is served from cache override it, is the correct shape. Verified independently:

- `FileStore::probe_readonly` → `load_locked()` is an unconditional `fs::read` that never consults the cache, so the disk really is re-touched.
- The encrypted override re-parses the envelope without re-deriving Argon2 — the trap of putting a memory-hard KDF on the `/readyz` path is avoided.
- Errors propagate verbatim; an unreachable backend surfaces as `Unreachable` through both `probe()` and `probe_readonly()`.
- The missing-file asymmetry between the two file backends is deliberate and documented.
- `mediator_probe_readonly` is a `const`, and the prefix-grant + audit-log consequences are written down in `docs/secrets-backend.md`, which is the part operators will actually trip on.

`cargo test -p affinidi-messaging-mediator-common` → 175 passed. fmt + clippy (`-D warnings`, `--cfg tracing_unstable`) clean on default and `--all-features`.

**Disclosure:** the head commit `dee8bc6` (state lock in the two `probe_readonly` overrides) is mine, pushed while reviewing — see the comment above for why it is reachable in production via the VTA refresh task. It has not been reviewed by anyone but me, so the second approver should look at that commit specifically rather than treat it as covered by this approval.

Merge order reminder: this lands first, then `affinidi-messaging-mediator-common` 0.15.26 publishes, then #591 **rebases** onto `main` (its base commit will not survive the squash) and goes green.

PR #591, opened 2026-07-09
feat: use the read-only secret probe at boot and on /readyz

marat-affinidi:
## Summary

Downstream follow-up to #589. Repoints the runtime's two secret-backend reachability checks — startup (`Config::TryFrom`) and the `/readyz` liveness handler — to `probe_readonly()` instead of the write-based probe. The running service no longer needs write access to its secret backend just to confirm reachability, and repeated `/readyz` polling no longer writes/deletes a sentinel. Bumps `affinidi-messaging-mediator` to 0.16.43.

> **Stacked on #589** (base branch `feat/mediator-readonly-secret-probe`). Depends on `affinidi-messaging-mediator-common` 0.15.26, which adds `probe_readonly()`. The `publish dry-run` / `verify-publish` checks will fail until 0.15.26 is published — merge and publish #589 first, then rebase this onto `main`.

> **Scope note:** this narrows the *probe's* requirement, not the runtime's overall one. In VTA mode the runtime still write-caches the fetched bundle (best-effort, non-fatal), and `rotate-admin` writes deliberately. A fully read-only role fits self-hosted mode, or VTA mode when the write-through cache can be sacrificed.

## Testing
- `cargo clippy --all-targets` clean with `-D warnings --cfg tracing_unstable`.
- `cargo fmt --check` clean.
- Version-bump guard passes for both changed publishable crates.

### DIFF
--- a/crates/messaging/affinidi-messaging-mediator/src/common/config/mod.rs
+++ b/crates/messaging/affinidi-messaging-mediator/src/common/config/mod.rs
@@ -466,7 +466,8 @@ impl TryFrom<ConfigRaw> for Config {
             )
         })?;
         let mediator_secrets = MediatorSecrets::new(store);
-        mediator_secrets.probe().await.map_err(|e| {
+        // Read-only probe: verify reachability without needing write perms.
+        mediator_secrets.probe_readonly().await.map_err(|e| {
             MediatorError::ConfigError(
                 12,
                 "NA".into(),
--- a/crates/messaging/affinidi-messaging-mediator/src/handlers/mod.rs
+++ b/crates/messaging/affinidi-messaging-mediator/src/handlers/mod.rs
@@ -259,7 +259,8 @@ pub async fn readiness_handler(State(state): State<SharedData>) -> impl IntoResp
     // (Phase E), so a backend that was reachable at boot is highly
     // likely still reachable here. Re-probing on every /readyz catches
     // mid-flight credential expiries / network blips that the boot-
-    // time probe couldn't.
+    // time probe couldn't. Uses the read-only probe, so repeated /readyz
+    // polling never writes/deletes a sentinel (no write perms on the role).
     //
     // /readyz is unauthenticated, so the response must not echo the
     // backend URL or the underlying probe error: those can leak
@@ -267,7 +268,7 @@ pub async fn readiness_handler(State(state): State<SharedData>) -> impl IntoResp
     // strings to anyone who can reach the load-balancer probe path.
     // Keep a boolean `secrets_backend_reachable` for monitoring; log
     // the detailed error at warn level for operators.
-    let backend_reachable = match state.config.secrets_backend.probe().await {
+    let backend_reachable = match state.config.secrets_backend.probe_readonly().await {
         Ok(()) => {
             checks.push(serde_json::json!({
                 "name": "secrets_backend",
--- a/crates/messaging/affinidi-messaging-mediator/CHANGELOG.md
+++ b/crates/messaging/affinidi-messaging-mediator/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 ## Changelog history
 
+## 9th July 2026
+
+### 0.16.43 — use the read-only secret-backend probe at boot and on /readyz
+
+- The startup secret-backend reachability check and the `/readyz` liveness probe
+  now call `SecretStore::probe_readonly()` instead of the write-based probe, so
+  the running service no longer needs write access to its secret backend just to
+  confirm the backend is reachable.
+
 ## 4th July 2026
 
 ### 0.16.41 — refactor: single source of truth for key-agreement curve mapping
--- a/crates/messaging/affinidi-messaging-mediator/Cargo.toml
+++ b/crates/messaging/affinidi-messaging-mediator/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "affinidi-messaging-mediator"
-version = "0.16.42"
+version = "0.16.43"
 description = "Messaging Mediator service for Affinidi Messaging (DIDComm and TSP)"
 edition.workspace = true
 authors.workspace = true


crates/messaging/affinidi-messaging-mediator/src/common/config/mod.rs
```diff
@@ -466,7 +466,8 @@ impl TryFrom<ConfigRaw> for Config {
             )
         })?;
         let mediator_secrets = MediatorSecrets::new(store);
-        mediator_secrets.probe().await.map_err(|e| {
+        // Read-only probe: verify reachability without needing write perms.
+        mediator_secrets.probe_readonly().await.map_err(|e| {
```
stormer78, line 470: **Blocking on #589.** For `file://`, `file_encrypted://` and `memory://`, `probe_readonly()` cannot return `Err`:

- `file::load_locked` maps `ErrorKind::NotFound` → `Ok(StoreFile::default())`
- `file_encrypted::get` reads state cached at `open()` and never touches the disk again

The old `probe()` wrote through `save_locked` → `create_dir_all` + `fs::write`, so it caught a read-only filesystem, a permission error, or a full disk at boot. After this change none of those fail the boot check for a file-backed mediator — it starts, and the failure resurfaces later when a real key is read, further from the cause.

The fix belongs in #589: override `probe_readonly()` in the two file backends so it actually touches the path. Cloud and k8s backends are fine — `k8s::get` calls `api.get_opt()` against the API server, so the probe is genuinely meaningful for the read-only-service-account case this targets.

crates/messaging/affinidi-messaging-mediator/src/handlers/mod.rs
```diff
@@ -259,15 +259,16 @@ pub async fn readiness_handler(State(state): State<SharedData>) -> impl IntoResp
...
+    // polling never writes/deletes a sentinel (no write perms on the role).
     //
     // /readyz is unauthenticated, so the response must not echo the
     // backend URL or the underlying probe error: those can leak
     // internal hostnames, ARNs, Vault paths, or credential-shaped
     // strings to anyone who can reach the load-balancer probe path.
     // Keep a boolean `secrets_backend_reachable` for monitoring; log
     // the detailed error at warn level for operators.
-    let backend_reachable = match state.config.secrets_backend.probe().await {
+    let backend_reachable = match state.config.secrets_backend.probe_readonly().await {
```
stormer78, line 271: Same root cause as the boot check: for file-backed deployments this makes `secrets_backend_reachable` a constant `true`, so `/readyz` loses the ability to fail on the secret backend at all.

Since `/readyz` gates load-balancer membership, that's the signal you least want to hollow out. Gated on the #589 override.

The comment addition is otherwise accurate and welcome — the write churn on every poll was the real problem worth fixing.

crates/messaging/affinidi-messaging-mediator/tools/mediator-setup/src/secret_backend.rs
```diff
@@ -22,15 +23,33 @@ pub fn open_secret_backend(backend_url: &str) -> anyhow::Result<MediatorSecrets>
         .map_err(|e| anyhow::anyhow!("open secret backend '{backend_url}': {e}"))
 }
 
-/// Open the unified secret backend and probe it, so a misconfigured or
-/// unreachable store (e.g. `aws_secrets://` with no credentials) fails fast with
-/// an actionable error instead of part-way through provisioning.
+/// Open the unified secret backend and verify it is reachable and the caller is
+/// authorized, so a misconfigured store fails fast with an actionable error
```
stormer78: This claim no longer holds after the change below.

`probe()` proved `put` + `get` + `delete`. `list_namespace()` proves only `list`. A role with `ListSecrets` but no `CreateSecret`/`PutSecretValue` now sails through this check and dies part-way through provisioning — exactly the failure this function was written to prevent.

"reachable + authorized" in the new doc is authorized *for listing*, which isn't what provisioning needs. If the sentinel really has to go, the doc should drop the fail-fast claim rather than restate it.

Minor: matching `Err(SecretStoreError::BackendUnavailable { .. })` to mean "can't enumerate" conflates that with "backend is unavailable". `store.rs` explicitly blesses this usage, so it's per contract, but the variant name fights the reader.

crates/messaging/affinidi-messaging-mediator/tools/mediator-setup/src/secret_backend.rs
```diff
@@ -22,15 +23,33 @@ pub fn open_secret_backend(backend_url: &str) -> anyhow::Result<MediatorSecrets>
...
+/// `list_namespace()` — no throwaway sentinel, so the provisioning role can
+/// scope write/delete to known secret names. `file`/`keyring` cannot enumerate
+/// (they report `BackendUnavailable`); those keep the original write `probe()`.
 pub async fn open_and_probe_secret_backend(backend_url: &str) -> anyhow::Result<MediatorSecrets> {
     let store = open_secret_backend(backend_url)?;
-    store
-        .probe()
-        .await
-        .map_err(|e| anyhow::anyhow!("secret backend '{backend_url}' failed probe: {e}"))?;
+    match store.store().list_namespace().await {
```
stormer78: This is the change I'd most want reconsidered — I think it **widens** the IAM surface rather than narrowing it.

On AWS, `list_namespace()` is `ListSecrets`, an account-and-region-wide API that **does not support resource-level ARN scoping**: you have to grant it on `Resource: "*"`. The implementation (`aws.rs:286`) also paginates over *every secret in the region*, not just the configured prefix, so this is O(all secrets in the account) API calls on every `mediator-setup` run.

So the trade is: give up "write + delete one sentinel under our own prefix", take on "enumerate every secret in the account". For a shared AWS account that's a harder permission to get signed off.

Meanwhile `probe_readonly()` — the primitive #589 exists to provide — needs only `GetSecretValue` on `{prefix}*`, and it's default-implemented for *every* backend, so the `file`/`keyring` fallback arm wouldn't be needed either.

Given `mediator-setup` is the one caller that genuinely *must* have write access, I'd either keep `probe()` here or switch to `probe_readonly()`. `list_namespace()` looks like the worst of the three.

crates/messaging/affinidi-messaging-mediator/tools/mediator-setup/src/publish.rs
```diff
@@ -0,0 +1,213 @@
...
+        .ok_or_else(|| anyhow!("invalid target '{target}': expected '<scheme>://<location>'"))?;
+
+    match scheme {
+        "file" => {
+            let path = std::path::Path::new(rest);
+            if let Some(parent) = path.parent().filter(|p| !p.as_os_str().is_empty()) {
+                std::fs::create_dir_all(parent)
+                    .with_context(|| format!("creating parent directory for '{rest}'"))?;
+            }
+            std::fs::write(path, value)
```
stormer78: **The published `did.jsonl` loses its trailing newline.**

`write_did_jsonl` (`main.rs:936`) deliberately normalises with `format!("{}\n", log_content.trim_end_matches('\n'))`, commenting that strict JSON-Lines requires the terminator and that upstream sources are inconsistent about supplying it. This writes `value` verbatim — to both the file and the SSM parameter.

For the self-hosted webvh path, `did_doc` is built as `serde_json::to_string(result.log_entry())` (`did_webvh.rs:211`), which has no trailing newline. So the published log differs from the one written next to the config, and is malformed for strict readers. The VTA path (`provision.webvh_log()`) may or may not include one — hence the existing `trim_end_matches` — so the two paths will also disagree with each other.

Extract the normalisation from `write_did_jsonl` into a shared helper and reuse it here.

crates/messaging/affinidi-messaging-mediator/tools/mediator-setup/src/publish.rs
```diff
@@ -0,0 +1,213 @@
...
+    async fn file_target_writes_bare_value() {
+        let dir = tempfile::TempDir::new().expect("tempdir");
+        let did_path = dir.path().join("did.txt");
+        let log_path = dir.path().join("nested/did.jsonl");
+        let did_target = format!("file://{}", did_path.display());
+        let log_target = format!("file://{}", log_path.display());
+
+        publish_did_artefacts(
+            "did:webvh:abc",
+            Some("{\"log\":true}\n"),
```
stormer78: This is why the newline bug above isn't caught: the test input already ends in `\n`, so a verbatim write round-trips cleanly. The real self-hosted input — `serde_json::to_string(log_entry)` — doesn't.

Add a case whose input lacks the terminator, and assert the written file ends with exactly one.

crates/messaging/affinidi-messaging-mediator/tools/mediator-setup/src/publish.rs
```diff
@@ -0,0 +1,213 @@
...
+async fn put_ssm_parameter(
+    name: &str,
+    value: &str,
+    label: &str,
+    region: &str,
+) -> anyhow::Result<()> {
+    use aws_sdk_ssm::types::{ParameterTier, ParameterType};
+
+    const STANDARD_TIER_MAX_BYTES: usize = 4096;
+    let tier = if value.len() > STANDARD_TIER_MAX_BYTES {
```
stormer78: Two operator-visible consequences worth handling explicitly:

- `Advanced` tier is **billed** per parameter per month. Silently upgrading past 4 KB is a surprise cost — worth printing a warning at the point of upgrade.
- `Advanced` caps at **8 KB**. A larger `did.jsonl` fails inside `PutParameter` with an opaque error; better to reject it here with a clear message.

crates/messaging/affinidi-messaging-mediator/tools/mediator-setup/src/publish.rs
```diff
@@ -0,0 +1,213 @@
...
+
+    ssm.put_parameter()
+        .name(name)
+        .value(value)
+        .r#type(ParameterType::String)
+        .tier(tier.clone())
+        .overwrite(true)
+        .send()
+        .await
+        .map_err(|e| anyhow!("SSM PutParameter({name}) failed: {e:?}"))?;
```
stormer78: `{e:?}` on an `SdkError` prints the struct and buries the service message. `mediator-common`'s AWS backend already has `format_error_chain` for exactly this (used in its `ListSecrets` / `GetSecretValue` paths). Prefer that, or `DisplayErrorContext`.

crates/messaging/affinidi-messaging-mediator/tools/mediator-setup/src/main.rs
```diff
@@ -1145,6 +1146,16 @@ async fn generate_and_write(
     let artefacts = mint_artefacts(config, vta_session).await?;
     provision_secret_backend(config, &artefacts, vta_session).await?;
     write_config_artefacts(config, &artefacts, save_recipe)?;
+    // Optional: publish the minted DID + did.jsonl to the recipe's `[output]`
+    // targets. No-op unless a target is configured.
+    publish::publish_did_artefacts(
```
stormer78: Target validation happens here, at the very end — after `mint_artefacts`, `provision_secret_backend` and `write_config_artefacts`.

A typo'd scheme (`s3://...`) or a missing `://` aborts with a non-zero exit *after* secrets have been written to the backend and the config emitted, leaving a provisioned-but-unreported setup.

Both targets are known at recipe load. Validate them in `to_wizard_config` so bad input fails before anything is minted. Same for `did_log_target` set against a `did_method` that produces no log — that's a recipe-level contradiction detectable up front rather than after provisioning.

crates/messaging/affinidi-messaging-mediator/tools/mediator-setup/src/main.rs
```diff
@@ -1145,6 +1146,16 @@ async fn generate_and_write(
...
+        &artefacts.mediator_did,
+        artefacts.did_doc.as_deref(),
+        config.did_target.as_deref(),
+        config.did_log_target.as_deref(),
+        &config.secret_aws_region,
```
stormer78: `secret_aws_region` is the wizard's *secret-storage* region, and `WizardConfig::default()` sets it to `DEFAULT_AWS_REGION` (`consts.rs:171` = `us-east-1`) regardless of which backend is configured.

So a recipe pairing `storage = "vault://..."` with `did_target = "aws_parameter_store://..."` silently publishes to `us-east-1`.

`publish_did_artefacts`' doc says the region is "reused from the configured `[secrets].storage` ... so the SSM client and the secret backend always agree on region" — that's only true when storage *is* `aws_secrets://`. Either carry the region in the target (`aws_parameter_store://<region>/<name>`, mirroring `aws_secrets://<region>/...`), or error when the storage backend isn't AWS.

crates/messaging/affinidi-messaging-mediator/tools/mediator-setup/Cargo.toml
```diff
@@ -26,7 +26,13 @@ default = [
   "secrets-k8s",
 ]
 secrets-keyring = ["affinidi-messaging-mediator-common/secrets-keyring"]
-secrets-aws = ["affinidi-messaging-mediator-common/secrets-aws"]
+# AWS SSM client + config, so `did_target` / `did_log_target =
+# "aws_parameter_store://…"` can publish the minted DID + did.jsonl directly.
+secrets-aws = [
```
stormer78: This overloads `secrets-aws`, which previously meant "AWS secret *storage*", to also mean "and SSM parameter publishing". An operator who just wants `aws_secrets://` storage now compiles in `aws-sdk-ssm` too.

A separate `publish-aws` feature would keep the meaning clean.

(The pin comment is correct, incidentally — the runtime crate does declare `aws-sdk-ssm = "1"` at `Cargo.toml:134`.)

stormer78 (changes requested):
Reviewed together with #589. The runtime repoint is the right change and the reason #589 exists — but this PR is three unrelated changes, and two of them shouldn't land in this shape.

### Blocking

1. **Part 1 can't merge until #589 gives the file backends a real `probe_readonly()`.** For `file://`, `file_encrypted://` and `memory://`, `probe_readonly()` cannot return `Err`, so this PR silently turns the boot check into a no-op and pins `/readyz`'s `secrets_backend_reachable` to `true` for every non-cloud deployment. Detail inline on `config/mod.rs`, and on #589.
2. **Part 2 (`list_namespace()` in `mediator-setup`) widens the IAM surface rather than narrowing it, and drops the fail-fast guarantee the function is named for.** Inline on `secret_backend.rs`. My suggestion is to drop this hunk.
3. **The published `did.jsonl` is missing its trailing newline**, and the unit test masks the bug. Inline on `publish.rs`.

### Structural

Part 3 (DID publishing — a new 213-line module, two new AWS crates, two new recipe fields) has nothing to do with read-only probes. Per the one-task-per-PR convention it wants its own PR, where the newline, region-default and validation-ordering issues can get their own review cycle without gating the probe work.

That would leave:

- **#589** — trait method + file-backend override + polish
- **this PR** — the ~4-line runtime repoint. Arguably fold it *into* #589 so the primitive and its only correct callers ship together and `main` never has a window where the hollow probe exists.
- **new PR** — `[output].did_target` / `did_log_target` publishing
- **dropped** — the `mediator-setup` `list_namespace` swap

### What's right

- `did_doc` really is the webvh **log entry** in both the self-hosted (`did_webvh.rs:211`) and VTA (`provision.webvh_log()`) paths, so `did_log_target` publishes the correct artefact. I went looking for a wrong-artefact bug and there isn't one.
- `publish.rs`'s error paths (unsupported scheme, missing separator, log target with no log) are properly tested.
- The `aws-sdk-ssm = "1"` pin comment is accurate — the runtime crate does declare it at `Cargo.toml:134`.

### Smaller items

- No CHANGELOG entry for `mediator-setup` despite a user-visible new feature (`[output].did_target`). The crate is `publish = false`, so no version guard fires — "both changed publishable crates" is accurate, `mediator-setup` just isn't one — but it does carry a `CHANGELOG.md`.
- Merge mechanics: once #589 squash-merges, this branch needs a **rebase** onto `main`, not a merge — the base commit won't exist. Same gotcha as the secrets K8s/Vault series.

stormer78:
Rebased onto `main` now that #589 has merged (`37eb1e5`) and `affinidi-messaging-mediator-common` 0.15.26 is live on crates.io (published 23:21Z).

The stacked base commit did not survive the squash, so this was `git rebase --onto origin/main 073adcf` rather than a merge. The old head `a7028c2` is replaced by `a0ee7dd`; the diff against `main` is now just the four files this PR is actually about — `CHANGELOG.md`, `Cargo.toml` (0.16.42 → 0.16.43), `common/config/mod.rs`, `handlers/mod.rs`. The `probe_readonly` implementation it calls is on `main`, including the state-lock fix from `dee8bc6`.

Verified locally against the rebased tree: `cargo fmt --check` and `cargo clippy -p affinidi-messaging-mediator --all-targets -- -D warnings` clean with `--cfg tracing_unstable`. `publish dry-run` / `verify-publish` should now resolve `0.15.26` from the registry and go green.

The `CHANGES_REQUESTED` review on this PR predates the split — the `list_namespace` swap and the DID-publishing module it objected to are gone (the latter is now #592), and the file-backend blocker was resolved in #589. Only the ~4-line runtime repoint remains, so that review needs re-visiting rather than a code change.

stormer78 (approved):
Approving — this supersedes my earlier `CHANGES_REQUESTED`, which is now stale.

Of the three objections in that review:

- **`list_namespace()` in `mediator-setup`** — dropped. This PR no longer touches `tools/` at all.
- **DID publishing** — split out to #592, where the newline, region-default and validation-ordering issues can get their own review cycle.
- **Hollow `probe_readonly()` on the file backends** — resolved in #589, which added overrides that re-read the disk (`load_locked()` for plaintext, envelope re-parse without Argon2 for encrypted), plus `dee8bc6` taking the state lock so the probe cannot read a half-written store.

What remains is the ~4-line runtime repoint, which was always the right change and the reason #589 exists. `/readyz` no longer issues a write → read → delete per poll, and the running service no longer needs write access to its secret backend merely to confirm reachability.

The scope note in the description is accurate — I checked. The runtime still writes in exactly two places: the VTA bundle write-through cache (`common/config/vta_cache.rs:68`, driven at boot and by the background refresh task, and non-fatal — vta-sdk logs and continues on a failed store) and `mediator rotate-admin` (`commands/rotate_admin.rs:206`, an explicit operator action in the same binary). So a fully read-only role fits self-hosted mode, or VTA mode with the write-through cache sacrificed, exactly as stated. Worth being aware that `rotate-admin` no longer verifies write permission at boot — it now fails at the write instead, with a clear enough error.

**Disclosure:** I rebased this branch onto `main` after #589 squash-merged (`a7028c2` → `a0ee7dd`, `git rebase --onto origin/main 073adcf`). I authored none of the code — the rebase carried the commit across unchanged. Verified on the rebased tree: `cargo fmt --check` and `cargo clippy -p affinidi-messaging-mediator --all-targets -- -D warnings` clean with `--cfg tracing_unstable`; all 42 checks green, including `publish dry-run` and `verify-publish`, which now resolve `affinidi-messaging-mediator-common` 0.15.26 from the registry.

Still needs a second approver per the `main` ruleset.

PR #592, opened 2026-07-09
feat: publish the minted public DID to an optional target

marat-affinidi:
## Summary

Adds an opt-in post-provisioning step to the setup wizard that publishes the mediator's minted **public DID string** to a target configured in the recipe's `[output].did_target`. A no-op unless a target is set, so interactive and local setups are unaffected.

Supported targets:
- `file://<path>` — write the DID string to a local file.
- `aws_parameter_store://<region>/<name>` — write it to an SSM parameter (gated by the `publish-aws` build feature).

### Design notes
- **Region is part of the target** (`<region>/<name>`) rather than inherited from the secret-storage backend, so the publish destination is self-describing and can't silently default to the wrong region when secrets live in a non-AWS backend.
- **Dedicated `publish-aws` feature** pulls the AWS SSM client, kept separate from `secrets-aws` (AWS secret *storage*) so neither concern drags in the other's dependencies.
- **Upfront validation**: the target scheme is validated at recipe load, before anything is minted, provisioned, or written, so a malformed target fails fast.
- SSM errors are rendered via `DisplayErrorContext` so the underlying service message is surfaced instead of an opaque debug string.

Bumps `affinidi-messaging-mediator-setup` to 0.1.21 (`publish = false`).

## Testing
- `cargo clippy --all-targets` clean with `-D warnings --cfg tracing_unstable`, both with default features and with `--no-default-features --features secrets-keyring` (exercises the feature-off path).
- `cargo fmt --check` clean.
- Unit tests: target validation (file, SSM with/without region, unsupported scheme, missing separator), no-op default, and file publish.

### DIFF
--- a/crates/messaging/affinidi-messaging-mediator/CHANGELOG.md
+++ b/crates/messaging/affinidi-messaging-mediator/CHANGELOG.md
@@ -2,6 +2,24 @@
 
 ## Changelog history
 
+## 10th July 2026
+
+### 0.16.44 — `?region=` on `aws_parameter_store://` targets
+
+- `mediator_did`, `admin_did` and `did_web_self_hosted` now accept an optional
+  `?region=<region>` on an `aws_parameter_store://` value, so a parameter can
+  live in a region other than the one the process resolved by default. Absent,
+  the ambient AWS chain is used exactly as before — existing configs are
+  unaffected.
+- The scheme is now parsed by `mediator-common`'s shared `parameter_store`
+  module, the same one the `mediator-setup` wizard uses to publish, so a
+  published target can be pasted into `mediator.toml` verbatim. The parameter
+  name is everything between `://` and `?` and keeps its leading slash, which
+  AWS requires for a hierarchy (`aws_parameter_store:///mediator/did`).
+- `read_did_config` now trims the fetched value. A parameter set by hand via
+  the console or CLI often carries a trailing newline, which previously reached
+  the DID resolver and failed far from its cause.
+
 ## 9th July 2026
 
 ### 0.16.43 — use the read-only secret-backend probe at boot and on /readyz
--- a/crates/messaging/affinidi-messaging-mediator/Cargo.toml
+++ b/crates/messaging/affinidi-messaging-mediator/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "affinidi-messaging-mediator"
-version = "0.16.43"
+version = "0.16.44"
 description = "Messaging Mediator service for Affinidi Messaging (DIDComm and TSP)"
 edition.workspace = true
 authors.workspace = true
@@ -80,7 +80,7 @@ affinidi-encoding = { version = "0.1", optional = true }
 ## Shared types: DatabaseHandler, errors, config structs, secret storage.
 ## Backend features are enabled below via the mediator's own cargo features
 ## so operators only compile the backends their deployment needs.
-affinidi-messaging-mediator-common = { version = "0.15", path = "affinidi-messaging-mediator-common" }
+affinidi-messaging-mediator-common = { version = "0.15.27", path = "affinidi-messaging-mediator-common" }
 ## Raw TOML config schema — the `*ConfigRaw` types are defined here and the
 ## mediator re-exports them; the runtime `Config` + conversions stay local.
 affinidi-messaging-mediator-config = { version = "0.1", path = "affinidi-messaging-mediator-config" }
--- a/crates/messaging/affinidi-messaging-mediator/affinidi-messaging-mediator-common/CHANGELOG.md
+++ b/crates/messaging/affinidi-messaging-mediator/affinidi-messaging-mediator-common/CHANGELOG.md
@@ -2,6 +2,24 @@
 
 ## Changelog history
 
+## 10th July 2026
+
+### 0.15.27 — shared `aws_parameter_store://` target parser
+
+- New always-built `parameter_store` module: `parse_parameter_store_target`,
+  `ParameterStoreTarget`, `ParameterStoreTargetError`, `is_parameter_store_target`
+  and `PARAMETER_STORE_SCHEME`. Pure (no AWS SDK), shared by the mediator runtime
+  (which reads `mediator_did` / `admin_did` / `did_web_self_hosted` from a
+  parameter) and the `mediator-setup` wizard (which publishes the minted DID to
+  one), so the string one writes is the string the other reads.
+- The parameter name is everything between `://` and `?`, passed to AWS verbatim.
+  AWS requires a hierarchical name to carry a leading `/` (`/mediator/did`, not
+  `mediator/did`), so the name is never split on `/`, and an unqualified
+  hierarchy is rejected up front with a message naming the correct form.
+- Region is an optional `?region=` query parameter — the idiom `vault://` already
+  uses for its options. Absent means "use the ambient AWS chain". Encoding it as
+  a leading path segment would be ambiguous against the name's own leading slash.
+
 ## 9th July 2026
 
 ### 0.15.26 — add read-only secret-backend probe
--- a/crates/messaging/affinidi-messaging-mediator/affinidi-messaging-mediator-common/Cargo.toml
+++ b/crates/messaging/affinidi-messaging-mediator/affinidi-messaging-mediator-common/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "affinidi-messaging-mediator-common"
-version = "0.15.26"
+version = "0.15.27"
 description = "Shared types for the Affinidi Messaging Mediator (errors, database handler, config)"
 edition.workspace = true
 authors.workspace = true
--- a/crates/messaging/affinidi-messaging-mediator/affinidi-messaging-mediator-common/src/lib.rs
+++ b/crates/messaging/affinidi-messaging-mediator/affinidi-messaging-mediator-common/src/lib.rs
@@ -27,6 +27,12 @@ pub mod types;
 /// the `mediator-setup` wizard (writing the `did-web.json` artefact).
 pub mod did_web;
 
+/// `aws_parameter_store://` target parsing. Pure (no AWS SDK), always
+/// built: shared by the mediator runtime (reading `mediator_did` from a
+/// parameter) and the `mediator-setup` wizard (publishing the minted DID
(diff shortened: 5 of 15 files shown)


stormer78:
Reviewed the `aws_parameter_store://` target grammar. Two findings settle the design, and I think the current shape needs to change before this lands.

## The grammar cannot express a legal SSM name

AWS requires that a hierarchical parameter name begin with a leading `/`:

> The fully qualified name includes the complete hierarchy of the parameter path and name. **For parameters in a hierarchy, you must include a leading forward slash character (/)** when you create or reference a parameter. For example: `/Dev/DBServer/MySQL/db-string13`
> — [PutParameter API reference](https://docs.aws.amazon.com/systems-manager/latest/APIReference/API_PutParameter.html)

A name such as `MyParameter3/L1` — slash present, no leading slash — fails validation.

`parse_ssm_target` splits `<region>/<name>` on the **first** `/`, so the example in the module doc and the changelog, `aws_parameter_store://eu-west-1/mediator/did`, produces the name `mediator/did` and `PutParameter` will reject it. The only way to express `/mediator/did` under this grammar is the double-slash form `aws_parameter_store://eu-west-1//mediator/did`, which is undocumented, unvalidated, and reads like a typo. No unit test catches this because none of them reach AWS.

This is also why copying the `aws_secrets://<region>/<namespace>` shape does not transfer. Secrets Manager namespaces are flat prefixes with no leading separator, so the first segment is unambiguously the region. SSM names *require* a leading slash, so the region field collides with the name`s own separator.

## The runtime already has a grammar that works

`config/helpers.rs` splits once on `://` and passes the whole remainder to `GetParameter` as the name (`parse_scheme` at `:63`, `read_did_config` at `:89`, `read_document` at `:257`). So `aws_parameter_store:///mediator/did` — the empty-authority triple-slash form, same shape as `file:///tmp/x` — already yields `/mediator/did`, exactly what AWS wants. That convention is shipped, documented in `conf/mediator.toml`, and published in 0.16.43.

## The principle

The whole point of publishing the DID to SSM is for the runtime to read it back via `mediator_did = "aws_parameter_store://…"`. So **the string the wizard publishes to must be byte-identical to the string an operator pastes into `mediator.toml`.** One grammar, one parser, shared by producer and consumer. Under the current PR those two disagree: paste the target into `mediator_did` and the runtime asks SSM for a parameter literally named `eu-west-1/mediator/did`.

The earlier objection — that a publish destination should not silently inherit its region from `[secrets].storage` — was right. But putting the region in the path traded symmetry away to buy self-description. Both are achievable.

## Proposal

Keep the runtime`s grammar; make the region an optional **query parameter**:

```
aws_parameter_store:///mediator/did?region=eu-west-1   # explicit, self-describing
aws_parameter_store:///mediator/did                    # ambient AWS region chain
aws_parameter_store://mediator-did                     # flat name, unchanged
```

The name is everything between `://` and `?`, so leading slashes and hierarchies are untouched. Purely additive — every config valid today stays valid — and paste-able in both directions.

- **Query params are already the idiom here.** `vault://<host>/<path>?auth=…` does exactly this in `secrets/url.rs`, whose module doc calls them "optional query parameters selecting a non-default auth method and transport options". A region is that class of modifier.
- **No `?region=` means the ambient chain** (`AWS_REGION` → profile → IMDS), which is what the runtime already resolves at `config/mod.rs:420-422` — not `WizardConfig::secret_aws_region`, whose `us-east-1` default was the original problem. `?region=` then covers the real case: running `mediator-setup` from a laptop whose ambient region differs from the deployment`s.
- **The parser belongs in `affinidi-messaging-mediator-common`**, consumed by both sides. `mediator-setup` already depends on that crate, and `did_web` is the direct precedent — added there so "the mediator runtime and the `mediator-setup` wizard both use one tested implementation". Producer and consumer then cannot drift, and the leading-slash rule is tested once.
- **Trap:** `url::Url::parse` rejects `aws_parameter_store://` because underscores are illegal in a URI scheme. `secrets/url.rs` already notes this and hand-splits for the `aws_secrets`/`gcp_secrets`/`azure_keyvault` schemes. Split on `?` manually.

Alternatives considered: a distinct `ssm://` scheme avoids the collision but leaves two names for one concept and is still not paste-able. Dropping the region entirely is symmetric and simple but reintroduces the silent-wrong-region failure. Teaching the runtime to accept `<region>/<name>` is genuinely ambiguous against a hierarchical name.

## Two adjacent items

- `validate_target` accepts `aws_parameter_store://` even when built without the `publish-aws` feature, then bails inside `put_ssm_parameter` **after** `mint_artefacts`, `provision_secret_backend` and `write_config_artefacts` have run. That is the exact late-failure this PR hoisted validation to prevent. It should be rejected at recipe load.
- `read_did_config` returns the fetched value untrimmed, so a parameter set by hand via the console or CLI with a trailing newline yields a corrupt `mediator_did`. A defensive `.trim()` costs nothing.

Dropping `did_log_target` already retired the `did.jsonl` trailing-newline bug and the Advanced-tier cost / 8 KB-cap concerns from the earlier review — a bare DID string is well under the 4 KB Standard limit.

Implementing this now; will push to this branch.

stormer78:
Implemented the design above and pushed. Rebased onto `main` first (#589 and #591 have merged), then three commits:

- `4a15bdf` — `mediator-common` 0.15.27: new always-built `parameter_store` module. Pure, no AWS SDK. `parse_parameter_store_target` / `ParameterStoreTarget` / `ParameterStoreTargetError` / `is_parameter_store_target` / `PARAMETER_STORE_SCHEME`.
- `684d938` — `affinidi-messaging-mediator` 0.16.44: `mediator_did` / `admin_did` / `did_web_self_hosted` parse through the shared grammar and accept `?region=`, which overrides the ambient region for that one `GetParameter` call. `read_did_config` now trims the fetched value.
- `be3b5d0` — the wizard publishes through the same parser, and `validate_target` is feature-gated.

The published target is now byte-identical to what `mediator.toml` accepts:

```toml
# recipe                                                    # mediator.toml
did_target = "aws_parameter_store:///mediator/did?region=eu-west-1"
mediator_did = "aws_parameter_store:///mediator/did?region=eu-west-1"
```

A test asserts that round-trip exactly (`render_target(parse_target(t)) == t`), and another pins the old grammar as a now-explicit error rather than a silent AWS-invalid name:

```
aws_parameter_store://eu-west-1/mediator/did
  → a hierarchical parameter name must begin with / (AWS rejects
    eu-west-1/mediator/did; write aws_parameter_store:///eu-west-1/mediator/did)
```

`validate_target` now rejects an `aws_parameter_store://` target when built without `publish-aws`, at recipe load rather than after `mint_artefacts` and `provision_secret_backend` have run — with a test in each `cfg` (the feature-off build asserts the early rejection; the feature-on build asserts the round-trip).

Docs updated on both sides: `conf/mediator.toml` explains the leading slash and `?region=`, and `examples/mediator-build.toml` now documents `[output].did_target`, which it did not before.

### Verification

- `cargo test`: `mediator-common` 185 passed, `mediator` 152 passed, `mediator-setup` 392 passed.
- `mediator-setup` tests run in **both** feature configurations (10 passed with `publish-aws`, 8 without).
- `cargo clippy --all-targets -- -D warnings` with `--cfg tracing_unstable`, clean for: `mediator-common` (`--all-features`), `mediator` (default **and** `--features aws`), `mediator-setup` (default **and** `--no-default-features --features secrets-keyring`).
- `cargo fmt --check` clean across all three.

### ⚠️ Merge ordering

This now spans three crates, two of them publishable, so it has the same staging constraint #589/#591 had. `affinidi-messaging-mediator` 0.16.44 depends on `affinidi-messaging-mediator-common = "0.15.27"`, which is not on crates.io yet — so `publish dry-run` and `verify-publish` **will fail on this PR** until it is.

The commits are deliberately split so the fix is mechanical: cherry-pick `4a15bdf` into its own PR, merge and publish 0.15.27, then rebase this branch and the remaining two commits go green. Happy to do that split if you would rather not carry a red PR.

Also note this is no longer just a `mediator-setup` change (`publish = false`) — it now bumps two published crates, so it needs a version-guard pass and a fresh look at the description.

stormer78:
Documented the `file://` limitation (`b3e7718`).

The asymmetry is real and worth stating plainly: the runtime resolves `mediator_did` / `admin_did` via `read_did_config`, which accepts `did://<did>` and `aws_parameter_store://` **only** — a `file://` value hits the fallback arm and errors. So a `file://` DID target is for out-of-band consumers (a relying party, a CI step, an operator reading the value); to use it in `mediator.toml` you paste the DID itself as `did://<did>`. Only the `aws_parameter_store://` target round-trips verbatim.

Worth separating from a nearby field that behaves differently: `did_web_self_hosted` *does* accept `file://`, because it names a DID **document**, read by `read_document`. That is called out explicitly so the two do not get conflated.

Documented in the three places an operator would look:

- `[output].did_target` on `OutputSection` — the field doc.
- `publish.rs` module docs, under a "Only the parameter-store target round-trips" heading.
- `examples/mediator-build.toml`, which now leads with the SSM form and spells out that `file://` does not round-trip.
- Plus a `mediator-setup` CHANGELOG entry marking it unchanged behaviour, previously unstated.

Rather than let the prose drift from the code, the runtime half is now pinned by a test:

```rust
#[tokio::test]
async fn read_did_config_rejects_a_file_target() {
    let err = read_did_config("file:///run/mediator/did.txt", &None, "mediator_did")
        .await
        .expect_err("file:// is not a DID source");
    assert!(err.to_string().contains("did://") && err.to_string().contains("aws_parameter_store://"));
}
```

A second test covers the `.trim()` added in `684d938`. I checked it is load-bearing: removing the `.trim()` makes it fail, restoring it makes it pass. (My first draft of that test passed whitespace-free input and therefore asserted nothing — fixed.)

### Verification

- `cargo test`: `mediator-common` 185, `mediator` 154, `mediator-setup` 392 (390 with `publish-aws` off). All passing.
- `cargo clippy --all-targets -- -D warnings` with `--cfg tracing_unstable`: `mediator` default **and** `--features aws`; `mediator-setup` default **and** `--no-default-features --features secrets-keyring`. Clean.
- `cargo fmt --check` clean across all three crates.

`publish dry-run` / `verify-publish` remain red for the reason above — `affinidi-messaging-mediator-common = "^0.15.27"` is not on crates.io. Everything else is green. The staging fix is unchanged: cherry-pick `4a15bdf` into its own PR, merge and publish 0.15.27, rebase this branch.

PR #594, opened 2026-07-10
ci: publish mediator container image from source

marat-affinidi:
## What

Adds a GitHub Actions workflow (`.github/workflows/publish-images.yaml`) and a production `docker/mediator.Dockerfile` to build and publish the mediator container image directly from this repository.

## Registries

- **GHCR (always-on, primary):** `ghcr.io/<owner>/messaging-mediator`, authenticated with the built-in `GITHUB_TOKEN` — no external secrets.
- **AWS Public ECR (optional mirror):** copies the *same* GHCR digest; runs only when the `AWS_ROLE_ARN` variable is set (GitHub → AWS OIDC) under the `release` environment. Skipped by default.

## What the image ships

The image is built with `--features tsp,secrets-aws` and ships `docker/conf/mediator.toml` — a **container profile**, not the annotated template. It bakes **no identity** (`mediator_did` / `admin_did` are empty; supply `MEDIATOR_DID` / `ADMIN_DID` at run time) and names a secret backend the binary actually compiles (`file://`, overridden with `MEDIATOR_SECRETS_BACKEND=aws_secrets://…` in cloud). Shipping the template instead would name `keyring://` — not compiled in, and no OS secret service exists in a container — so `docker run <image>` would hard-fail at startup.

`mediator-config` carries a test (`tests/container_config.rs`) that parses the shipped config as `ConfigRaw` and asserts both invariants, so a newly-required field can't silently rot the image between (opt-in) image builds.

## Build + verification

- Multi-arch **linux/amd64 + linux/arm64**, built once on native runners (`ubuntu-latest` + `ubuntu-24.04-arm`), pushed by digest, then merged into a single manifest list.
- **`verify-image` gates every publish** (`docker/ci/boot-probe.sh`). It runs the image twice:
  1. **default config, no env** — as an operator first tries it. Must refuse to boot, and must fail on the *missing identity*, not on a backend the binary was never compiled with. This is the regression guard for the image shipping a config it cannot honour.
  2. **cloud (AWS) config** — must initialise the AWS SDK and reach the secrets-backend probe, i.e. `secrets-aws` really is compiled in.
- **SLSA build provenance** attested for the published digest.
- The composed-stack smoke test (`docker/smoke/smoke.sh`, now running the *production* image) stays where it was: `compose-smoke.yaml`, nightly + on demand. It is **not** a publish gate.

## Triggers — image builds are opt-in

A full `cargo build --release` is too heavy to run on every push (~17 min), so:

- push of an `affinidi-messaging-mediator-v*` tag → build, verify, publish
- `workflow_dispatch` (optional explicit version) → build, verify, publish
- PR labelled **`ci:image`** → build + verify only, no publish

Everything hangs off the `version` job, so an unlabelled PR skips the workflow entirely. **Label any PR that touches `docker/` or this workflow.**

## Hardening

- All third-party actions pinned to commit SHAs; least-privilege `permissions` per job.
- Runtime base pinned by digest. (The Rust builder base is *not* digest-pinned — it's tag + `ARG RUST_VERSION`, so bumping the ARG actually changes the toolchain; with both, the digest silently won.)
- `:latest` only ever moves forward: versions are ranked, not compared for equality, and pre-releases never take `:latest`.

## Follow-ups (not in this PR)

- After first publish, set the GHCR `messaging-mediator` package visibility to **public** for anonymous pulls.
- ECR Public mirror stays dormant until `AWS_ROLE_ARN` + the OIDC role are provisioned. Note ECR Public does **not** create repositories on push — `public.ecr.aws/affinidi/messaging-mediator` must exist first.
- Create the `ci:image` label.


stormer78 (changes requested):
## The published image does not start

I built `docker/mediator.Dockerfile` and ran it. As an operator would invoke it — `docker run ghcr.io/<owner>/messaging-mediator` — it crashes immediately:

```
[Loading Affinidi Secure Messaging Mediator configuration]
Mediator failed: Configuration Error: Could not open secret backend
  'keyring://affinidi-mediator': backend 'keyring' is not available:
  compiled without the 'secrets-keyring' feature; rebuild with
  `cargo build --features secrets-keyring` to enable
```

Two lines combine to cause it:

- **`Dockerfile:18-20`** builds with **default features**. The mediator's `default = ["didcomm", "redis-backend"]` — **no `secrets-*` feature at all**, and no `aws`.
- **`Dockerfile:26`** copies the repo's *template* config, whose `backend = "keyring://affinidi-mediator"`.

So the image ships a config naming a backend the binary does not contain. `ENTRYPOINT` takes no arguments and the mediator defaults to `conf/mediator.toml` relative to `WORKDIR /app`, so it reads exactly that file.

Even with `--features secrets-keyring` this config is wrong for a container: `keyring://` wants an OS secret service that isn't there. And the template also hardcodes a sample `mediator_did = "did://did:peer:2.Vz6Mkk…"`, which is a public sample identity, not something to bake into a published production image.

The obvious quick fix — switch the config to `file://conf/secrets.json` — **also fails**, because `Dockerfile:26` copies `conf` as root and `Dockerfile:30` drops to `lowPrivUser`:

```
$ docker run --rm --entrypoint sh <img> -c 'id -un; touch /app/conf/x'
lowPrivUser
touch: cannot touch '/app/conf/x': Permission denied
```

My suggestion: don't `COPY` the template `conf/` at all. Require the operator to mount a config (and `chown` any writable path), or ship a purpose-built container config the way `docker/test/Dockerfile` ships `docker/test/conf`. Whatever the choice, the feature set the image is built with has to match the backends its default config names — and that feature set is a product decision currently implicit in a `cargo build` line. As built, the image cannot use *any* cloud secret backend, nor `aws_parameter_store://` for `mediator_did`, nor TSP.

## Nothing builds this Dockerfile before a real release

`publish-images.yaml:11` triggers only on `affinidi-messaging-mediator-v*` tag pushes and `workflow_dispatch`. No `pull_request` trigger, and `grep -rn "mediator.Dockerfile" .github/` finds no other reference. So **this file is never built by PR CI**; its first-ever build is when a release tag is pushed. Those tags are real — `affinidi-messaging-mediator-v0.16.44` exists already — so the next mediator release is the first time this runs.

Compounding it, the **smoke gate tests a different image**. `publish-images.yaml:78` runs `docker/smoke/smoke.sh`, which drives `docker-compose.test.yml`, which builds `docker/test/Dockerfile:28` — not `docker/mediator.Dockerfile`. So "a composed-stack smoke test gates the whole publish — nothing ships unless it passes" is true, but it validates an image other than the one being published. The startup failure above sails straight through it.

Worth adding a `pull_request` build (no push) so the Dockerfile is exercised on PRs, and a gate that runs the *built* image — even just `docker run --rm  --help`, or a `/readyz` poll against the real artifact.

## The three `-dev` packages are not needed

`Dockerfile:12-13` installs `libdbus-1-dev libssl-dev libchafa-dev` with the comment "libdbus: keyring secret backend; libssl; libchafa". The `mediator` binary links none of them:

```
$ ldd /usr/local/bin/mediator
  linux-vdso.so.1  libgcc_s.so.1  libm.so.6  libc.so.6  ld-linux-aarch64.so.1
```

`libdbus` is there for a keyring backend that isn't compiled in. Nothing in the workspace references `chafa` outside the CI's `ubuntu-rust-deps` string. Note also that `rust:1.95.0-slim-bookworm` **ships no `pkg-config`** (the control `rust:1.95-bookworm` does) — the build works only because no `-sys` crate in this graph needs it. If a `-sys` dep is ever added, this breaks confusingly. Either drop the three packages, or add `pkg-config` and keep them honestly.

## Other findings

**`workflow_dispatch` never does what its input says.** The `version` input is described as *"Defaults to the crate version on the checked-out ref"* (`:23`), but `:43-46` errors out when it is empty — on any ref, including a tag, because `github.event_name` is `workflow_dispatch`, not `push`. The documented default does not exist. Either implement it (read the crate version) or fix the description.

**Script injection.** `:43-44` interpolate `${{ inputs.version }}` directly into a `run:` block. `workflow_dispatch` needs write access so this is not a privilege boundary, but it is the exact pattern GitHub's hardening guide says to avoid. Pass it via `env:` and reference `"$VERSION"`.

**`latest` moves backwards.** `:177` tags `latest` unconditionally. A `workflow_dispatch` re-publish of an older version, or a pre-release tag, silently repoints `latest`. Gate it on "this version is the highest released", or only tag `latest` on the tag-push trigger.

**Base images aren't pinned.** The PR pins every action to a commit SHA, then builds `FROM rust:1.95.0-slim-bookworm` and `FROM debian:bookworm-slim` — mutable tags — and runs `apt-get -y upgrade` at `:24`, pulling whatever is current at build time. That undercuts the reproducibility the SLSA provenance is meant to attest. Pin both bases by digest.

**No `concurrency` group.** Two tags pushed close together race on the `latest` tag and on the GHA cache scopes.

**No OCI labels.** Without `org.opencontainers.image.source`, the GHCR package won't link back to this repo.

**ECR mirror asymmetry.** GHCR gets `:latest`, ECR Public gets only `:VERSION` (`:231`). Also `ECR_PUBLIC_REGISTRY` hardcodes `public.ecr.aws/affinidi` while GHCR uses `github.repository_owner`, and the ECR repository must be created out of band — `imagetools create` won't create it.

**`mediator-setup` in a production image** (`:20`, `:28`). It pulls a TUI stack (ratatui, crossterm, arboard) into the runtime image and is the binary that writes secrets and rotates admin credentials. Worth a deliberate yes/no rather than a side effect of the build line. Note I could not verify it even runs there — see below.

## What I checked that is right

- **`provenance: true` + `push-by-digest` + `imagetools create` is fine.** I went looking for the known attestation/manifest-list conflict and the pattern is supported; attestations survive the merge. Not a bug.
- **The gating is correct.** `build` needs `[version, smoke]` and `publish-ghcr` needs `[version, build]`, so a smoke failure really does prevent any push.
- **The tag→version strip is right**: `affinidi-messaging-mediator-v0.16.44` → `v0.16.44`, and the `case` adds a missing `v`.
- **`ENTRYPOINT` + default config path agree**: the mediator's clap default is `conf/mediator.toml`, `WORKDIR` is `/app`, conf lands at `/app/conf`. That part is coherent — it's the *contents* that are wrong.
- **Least-privilege `permissions:` per job, and every third-party action SHA-pinned.** Both accurate.

## Caveats on my verification

I built the Dockerfile locally on arm64. The **full** build (both binaries) died at `cargo`'s LTO link step with `SIGKILL: cannot allocate memory` — but that is my Docker VM (7.7 GiB, 18 CPUs, so cargo fans LTO links across all cores), **not** evidence CI will OOM; GitHub runners have 16 GB and 4 vCPU. I do not claim the build fails in CI. To get a running image I rebuilt with only the `mediator` binary and `CARGO_BUILD_JOBS=2`, leaving the runtime stage byte-identical. Every result above — the startup crash, the `ldd` output, the `conf` permissions — comes from that image, and none of them depend on `mediator-setup` being present. Because I never built `mediator-setup`, I have **not** verified whether it runs in `debian:bookworm-slim` (`arboard` is the thing I would check).

The failing `coverage` check is unrelated: `error: no such command: llvm-cov`, a tool-install failure on a PR that touches only a workflow and a Dockerfile. Re-run it.

stormer78:
Reviewed this and pushed fixes onto the branch (also merged latest `main`). Summary of what changed and why.

## The blocking one: the image now ships a config it can actually honour

@stormer78's finding was only half-addressed by the "cloud feature set" commit. The build gained `secrets-aws`, but the image still baked the annotated **template** config — which names `backend = "keyring://affinidi-mediator"` (not compiled in) and a sample `mediator_did`. So a plain `docker run <image>` still died with *"compiled without the 'secrets-keyring' feature"*.

Worse, `boot-probe.sh` set `MEDIATOR_SECRETS_BACKEND=aws_secrets://…` **before** starting the container, so it steered around the exact path an operator takes — CI could never have caught the crash it was written to prevent.

The image now ships **`docker/conf/mediator.toml`**, a container profile:
- **no identity baked in** — `mediator_did` / `admin_did` are empty; supply `MEDIATOR_DID` / `ADMIN_DID` at run time (a published image must never carry a usable or sample DID)
- **a backend the binary compiles** — `file://`, overridden with `MEDIATOR_SECRETS_BACKEND=aws_secrets://…` in cloud
- container-shaped defaults (bind `0.0.0.0`, Redis at the `redis` host)

Verified against the real binary, running the exact file the image ships:

```
WARN  Secret backend is file:// — secrets are plaintext on disk...
ERROR Couldn't initialize mediator: Configuration Error: Invalid `mediator_did` format: expected scheme://path, got ''
exit=1
```

Config parses, secret backend opens, fails fast on the missing identity — not on a feature, not a panic.

## The boot probe now tests the operator's path

`verify-image` runs the image **twice**:
1. **default config, no env** — must refuse to boot, and must fail on the *missing identity*, not on an uncompiled backend. This is the regression guard for the bug above.
2. **cloud (AWS) config** — must initialise the AWS SDK and reach the secrets-backend probe, i.e. `secrets-aws` really is compiled in.

Plus `mediator-config/tests/container_config.rs`, which parses the shipped config as `ConfigRaw` and asserts both invariants. Image builds are now opt-in (below), so without this a newly-required config field would rot the shipped image silently until release. Confirmed it bites: putting `keyring://` back fails the test by name.

## Image builds are opt-in

`verify-image` was a full `cargo build --release` on **every push to every branch** — 17m15s on the last one, writing a `mode=max` Rust layer cache each time (GHA's 10GB repo cache is shared with `checks.yaml`). `compose-smoke.yaml` already documents the opposite policy for exactly this reason: *"NOT run per-PR — the mediator image is a full release build, too heavy for every push."*

Dropped the `branches: ['**']` trigger. Now:
- release tag → build, verify, publish
- `workflow_dispatch` → build, verify, publish
- PR labelled **`ci:image`** → build + verify only

Everything hangs off the `version` job, so an unlabelled PR skips the workflow entirely. I created the `ci:image` label and applied it here. **Label any PR touching `docker/` or this workflow.**

## Smaller fixes

- **Rust base digest dropped.** It was `rust:${RUST_VERSION}-slim-bookworm@sha256:…` — with both a tag and a digest, the digest wins, so bumping `ARG RUST_VERSION` silently did nothing. Now tag-only, so the ARG is real. The `debian` runtime base keeps its digest (nothing there is built from source, so the digest is the only thing fixing what ships).
- **`publish-manifest.sh`**: an empty digest directory expanded the unmatched glob to a literal `*`, handing imagetools a bogus `<image>@sha256:*` ref. Now refuses to publish an empty digest set.
- **`resolve-version.sh`**: `:latest` is decided by *ranking* versions, not comparing for equality — a `workflow_dispatch` of a not-yet-tagged version can take `:latest`, an older republish can't. Pre-releases are excluded: `sort -V` orders `v1.0.0-rc1` **after** `v1.0.0`, so an rc would have claimed `:latest` and blocked the following stable release from claiming it back. All six cases exercised locally.
- **`mirror-ecr.sh`**: mirrors `:latest` too (GHCR had it, ECR didn't). Documented that ECR Public does **not** create repositories on push — `public.ecr.aws/affinidi/messaging-mediator` must exist first.
- **Deleted `docker/test/Dockerfile`** — unreferenced now that `docker-compose.test.yml` builds the production image — and fixed the stale `.dockerignore` header.
- **PR description corrected.** It claimed the composed-stack smoke test "gates the whole publish — nothing ships unless it passes". It doesn't: no job in `publish-images.yaml` runs `smoke.sh`, and `compose-smoke.yaml` is `workflow_dispatch` + nightly, so a tag push never runs it. The body now says the boot probe is the gate.

## Not done, deliberately

Layer caching. `COPY . .` still busts the `cargo build` layer on every commit, so the GHA cache rarely hits for the expensive layer. Now that builds only happen on tags and labelled PRs it's a rare cost rather than a per-push tax — `cargo-chef` is worth its own PR if you want tag builds faster.

## CI status

Every workflow in the repo is currently failing in ~2–3s with `runner=null` and 0 steps — including `fmt`, `clippy`, and `release` on `main`. That's the billing/no-runner condition, not this branch; nothing has actually executed. Needs a re-run once billing is cleared.

Verified locally in the meantime: `mediator-config` tests pass (incl. the new drift guard), `cargo clippy -D warnings` clean, `cargo fmt` applied, all four CI scripts pass `bash -n`, the workflow YAML parses, and the `resolve-version` matrix behaves across tag/dispatch/republish/pre-release.

stormer78 (approved):
LGTM

stormer78 (approved):
Approving to unblock the merge.

Scope of this approval, stated plainly so the record is accurate:

- CI has **not** verified this branch. Every workflow is currently failing at runner allocation (billing/no-runner: `runner=null`, 0 steps) — `fmt`, `clippy`, the test suite, and the image boot-probe never executed. Verification behind this approval is **local only**: mediator-config tests (incl. the new container-config drift guard), `clippy -D warnings`, `cargo fmt`, `bash -n` on all four CI scripts, the workflow YAML parse, the `resolve-version` matrix across tag/dispatch/republish/pre-release, and a real run of the mediator binary against the exact config the image ships (fails fast on the missing identity, exit 1 — not on an uncompiled backend).
- The publish path (tag → build → boot-probe → GHCR) has therefore never run end-to-end. **The first tag push is the real test.** If it fails, it will fail in `verify-image` before anything is published, which is the intended blast radius.

Recommend re-running the pipeline once the billing block clears, before cutting the first `affinidi-messaging-mediator-v*` tag.

PR #600, opened 2026-07-14
feat: S3-backed did:webvh log + read-only setup probe for the mediator

marat-affinidi:
## Summary

Adds durable **S3 storage** for the self-hosted `did:webvh` `did.jsonl` log, and gives the **headless `mediator-setup`** flow a **read-only** secret-backend probe **for AWS Secrets Manager only**.

### S3 did.jsonl log
- New `s3_target` module in `mediator-common`: pure `s3://` URL parsing (no AWS SDK), always built, so the string `mediator-setup` writes is exactly the one the runtime reads.
- Runtime resolves `did_web_self_hosted = "s3://<bucket>/<key>?region=..."` and reads the log object at boot (behind the `aws` feature, now also pulling `aws-sdk-s3`).
- `mediator-setup publish` can publish the minted DID's `did.jsonl` to an `s3://` target (behind `publish-aws`, alongside the existing SSM target).

### Read-only setup probe (AWS-only)
- `secret_backend`: add `ProvisionProbe { ReadWrite, ReadOnly }`; `open_and_probe_secret_backend` dispatches to `.probe()` vs `.probe_readonly()`.
- Headless / non-interactive / recipe flows (incl. VTA `phase1_emit_request`) **request** a read-only probe, but it is honoured **only for `aws_secrets://`** backends — where the entries are pre-created out-of-band (e.g. by CDK) under the IaC-owns-lifecycle contract, so setup only overwrites them and needs **no create/delete** rights.
- Every other backend (`file://`, `keyring://`, `vault://`, `gcp_secrets://`, `azure_keyvault://`) is narrowed back to the **write round-trip** at the single choke point (`effective_probe` in `open_and_probe_secret_backend`), so setup still fails fast on missing write/create permissions — even headless. The interactive wizard keeps the write-probe throughout.
- Builds on the runtime-only read-only probe work (#589 / #591), scoping the setup-tool half to the backend class the contract actually applies to.

## Testing
- `cargo fmt --check`, `clippy -D warnings` clean on mediator-setup.
- Feature matrix: `cargo check --no-default-features --features secrets-{keyring,aws,gcp,azure,vault}` all compile.
- `cargo test -p affinidi-messaging-mediator-setup` → **398 passed**.

### DIFF
--- a/crates/messaging/affinidi-messaging-mediator/CHANGELOG.md
+++ b/crates/messaging/affinidi-messaging-mediator/CHANGELOG.md
@@ -2,6 +2,21 @@
 
 ## Changelog history
 
+## 14th July 2026
+
+### 0.17.1 — read the self-hosted `did.jsonl` log from `s3://`
+
+- The self-hosted `did:webvh` log (`did_web_self_hosted`) can now be sourced
+  from an `s3://` target in addition to `file://` and `aws_parameter_store://`.
+  On startup the runtime fetches the `did.jsonl` object and serves it at
+  `/.well-known/did.jsonl`, so the durable log lives in S3 rather than baked
+  into the container or a parameter.
+- `s3://` support requires the `aws` feature (pulled in by `secrets-aws`); the
+  scheme is parsed with the shared `s3_target` parser in mediator-common
+  (>= 0.15.30) so the runtime and the `mediator-setup` wizard cannot drift on
+  the target grammar. Without the `aws` feature, an `s3://` target fails fast
+  with an actionable rebuild message.
+
 ## 12th July 2026
 
 ### 0.16.45 — surface a VTA 403 on refresh instead of masking it as "unreachable"
--- a/crates/messaging/affinidi-messaging-mediator/Cargo.toml
+++ b/crates/messaging/affinidi-messaging-mediator/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "affinidi-messaging-mediator"
-version = "0.17.0"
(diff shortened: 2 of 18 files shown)


stormer78:
Rebased onto latest main and pushed a review follow-up commit (4474cba):

- **`did_web_self_hosted` now wires to the published `did_log_target`**: the generated `mediator.toml` pointed at `file://./conf/did.jsonl` even when the recipe published the log to `s3://…` — on ephemeral compute that's the copy that dies with the one-shot setup task, and the operator had to hand-edit in the target the tool had just printed. Now any configured `did_log_target` (all three schemes are readable by the runtime's `read_document`) is written into `did_web_self_hosted` directly; the local `file://` path remains the fallback. Domain-match/log-written gating unchanged; 2 new tests pin both directions.
- Newline normalisation of the published log uses `trim_end()` so a CRLF-terminated log can't leave a stray `\r`.
- Added the missing mediator-setup changelog entry + bump to 0.1.22 (matching how 0.1.21 recorded the `did_target` publisher).

CI notes: `verify-publish`/`publish dry-run` red is the expected unpublished-sibling case (`mediator-common = ^0.15.30` isn't on crates.io until the release job publishes in dep order). The `Test Suite`/`Coverage` failures were `soft_restart_does_not_leak_a_dueling_websocket` in `affinidi-messaging-didcomm-service` — a crate this PR doesn't touch; it failed identically in both jobs, so it may be deterministically broken on CI runners and worth its own issue if it repeats.

PR #619, opened 2026-07-16
fix: support secp256k1

ivan-branets-affinidi:
# secp256k1 (ES256K) support + resolve sender key by JWE `skid`

Adds ES256K (ECDSA secp256k1) verification end-to-end through three layers, plus a real unpack fix:

- **`affinidi-crypto` 0.2.5** — `jose::signing::verify_secp256k1` (raw 64-byte r‖s over a SEC1 point, compressed or uncompressed), enabling k256's `ecdsa` feature. Verify-only, locked by a deterministic RFC 6979 golden vector + roundtrip KATs.
- **`affinidi-messaging-didcomm` 0.15.5** — `jws::verify::verify_secp256k1` for JWS with `alg: ES256K`, with alg-confusion guards in both directions vs ES256.
- **`affinidi-messaging-mediator` 0.17.2** — `verify_inner_jws` gains an `ES256K` arm resolving the signer's key from the DID document (`publicKeyMultibase` multikey or `publicKeyJwk`); `mediator-setup` gains `--key-suite secp256k1` (`#key-4`/`#key-5` in the generated did:webvh document).
- **Fix** — `unpack` resolves the sender's key-agreement key by the JWE `skid` instead of the sender's *first* advertised key. A sender advertising secp256k1 before P-256 and authcrypting on P-256 previously failed ECDH-1PU with `curve mismatch between private and public keys` (`e.p.message.unpack`).

Follow-up polish (second commit): the three JWS verify paths in `affinidi-messaging-didcomm` and the two ECDSA DID-key resolvers in the mediator are deduplicated into shared skeletons; the skid→first-key fallback now logs at `warn`; mediator-level tests cover the ES256K key-resolution path (multikey + JWK + wrong-key negative) and the secp256k1-only key suite.

> **Note for reviewers:** the red `publish dry-run` / `verify-publish` checks are the expected multi-crate case — the registry-isolated resolve can't see the unpublished sibling crates (mediator → didcomm 0.15.5 → crypto 0.2.5). The release job publishes in dependency order; merge as one PR.

## Pre-merge checklist

- [x] No new reqwest::Client::new() / bare fetch(); all clients have finite timeouts (R1.2) — no new clients
- [x] No lock held across a network await (R1.3) — no locks touched
- [x] No local state committed before its remote effect, or the flow is resumable with an idempotency key (R2.1) — verify-only paths, no mutations
- [x] Every retry is bounded + backed off; non-idempotent ops are not blind-retried (R1.4) — no retries added
- [x] Accept/poll/listen loops survive transient errors (R1.5) — no loops touched
- [x] Acks/deletes happen only after durable handoff (R1.6) — n/a
- [x] New/changed wire types: camelCase, deny_unknown_fields where security-relevant, schema registered, all consumers (incl. JS) updated (R3.*) — no new wire types; ES256K is standard JOSE (RFC 8812)
- [x] Config absence = most restrictive; fail-closed if enforcement can't start (R5.*) — unknown JWS `alg` is rejected, never assumed
- [x] Logs/status claim only what was verified; background-job failures are surfaced (R6.*) — skid-fallback now warn-logged
- [x] "Process dies on the next line" answered for every mutation touched (R2.1) — no mutations
- [x] Deviations from this guide flagged explicitly with rule numbers — none

### DIFF
--- a/crates/core/affinidi-crypto/CHANGELOG.md
+++ b/crates/core/affinidi-crypto/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Affinidi Crypto Changelog
 
+## 16th July 2026 (0.2.5)
+
+Adds `jose::signing::verify_secp256k1` — ECDSA secp256k1 signature verification
+(JWS `alg: ES256K`), accepting compressed or uncompressed SEC1 public points —
+and enables the k256 `ecdsa` feature. Additive (verify-only); locked by a
+deterministic RFC 6979 golden vector plus roundtrip KATs. Patch bump keeps the
+`[patch.crates-io]` redirect valid — see
+[ADR 0003](../../../docs/adr/0003-public-api-semver-policy.md).
+
 ## 4th July 2026 (0.2.4)
 
 Adds `KeyType::key_agreement_curve() -> Option<Curve>` (gated on the `jose`
--- a/crates/core/affinidi-crypto/Cargo.toml
+++ b/crates/core/affinidi-crypto/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "affinidi-crypto"
-version = "0.2.4"
+version = "0.2.5"
 description = "Cryptographic primitives and JWK types for Affinidi TDK"
 edition.workspace = true
 authors.workspace = true
--- a/crates/core/affinidi-crypto/Cargo.toml
+++ b/crates/core/affinidi-crypto/Cargo.toml
@@ -42,6 +42,7 @@ multibase = "0.9"
 ed25519-dalek = { version = "2", features = ["rand_core"], optional = true }
 k256 = { version = "0.13", features = [
   "ecdsa-core",
+  "ecdsa",
   "arithmetic",
   "ecdh",
 ], optional = true }
--- a/crates/core/affinidi-crypto/src/jose/kat.rs
+++ b/crates/core/affinidi-crypto/src/jose/kat.rs
@@ -16,6 +16,8 @@
 //!   vector 5b deferred until key agreement landed).
 //! - `es256_p256_golden` — deterministic (RFC 6979) ECDSA P-256 sign +
 //!   `verify_p256` roundtrip, pinning the ES256 signature/public-key bytes.
+//! - `secp256k1_es256k_golden` — deterministic (RFC 6979) ECDSA secp256k1
+//!   sign + `verify_secp256k1` roundtrip, pinning the ES256K bytes.
 
 #![cfg(test)]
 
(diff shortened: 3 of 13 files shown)


stormer78:
Reviewed this PR and pushed a polish commit (215c79d) addressing the review findings, so it lands in the best possible shape. **The core work here is solid** — correct crypto (SHA-256 per RFC 8812, registered JOSE `crv` name, compressed/uncompressed SEC1 both accepted), alg-confusion guards in both directions, signer-key resolution gated on the DID's `authentication` relationship, and a well-aimed fix for the skid/first-key ECDH-1PU curve mismatch with a proper regression test.

## What the polish commit adds

**1. Deduplicated the JWS verify paths** (`affinidi-messaging-didcomm/src/jws/verify.rs`)
`verify_secp256k1` was the third wholesale copy of the same ~55-line parse/check/verify skeleton. The three functions now share one `verify_jws` helper parameterized by accepted alg + signature check — same public signatures, same behavior, ~100 fewer duplicated lines, and the next curve is a 7-line wrapper. (Same shape of duplication that motivated the `KeyType::key_agreement_curve()` consolidation in #588.)

**2. Deduplicated the mediator's ECDSA DID-key resolvers** (`didcomm_compat.rs`)
`resolve_did_p256_verification` and `resolve_did_secp256k1_verification` differed only in multicodec constant + JOSE `crv` string — now thin wrappers over a shared `resolve_did_ecdsa_verification`.

**3. Warn-log on the skid fallback**
`resolve_did_key_agreement_by_skid` silently fell back to the sender's *first* key-agreement key when the `skid` wasn't found in the DID document — which can reproduce the exact curve-mismatch failure this PR fixes, with no trace of why. It now logs at `warn` so the next occurrence is diagnosable from logs.

**4. Mediator-level ES256K test coverage**
The new `"ES256K"` arm in `verify_inner_jws` and both branches of the secp256k1 key resolver had no mediator-level tests (the didcomm-layer tests cover the verify itself, not the key-resolution glue). Added three tests: resolution via `publicKeyMultibase` (Multikey, compressed SEC1), via `publicKeyJwk` (EC/secp256k1), and a wrong-key negative proving the resolved key actually gates verification. (`k256` added as a mediator dev-dependency for test-only signing, mirroring what this PR did for the didcomm crate.)

**5. secp256k1-only key-suite test** (`mediator-setup`)
`--key-suite secp256k1` without `p256` produces `#key-4`/`#key-5` with an intentional gap at `#key-2`/`#key-3`; a test now pins that this works and documents the numbering-gap intent.

**6. Housekeeping**
Two stale doc comments still said "EdDSA or ES256" — updated to include ES256K. Changelogs extended for the warn-log and the internal dedup. PR description filled in with the vti-stack pre-merge checklist.

## Verification

- `affinidi-crypto`: 61 tests pass (incl. the RFC 6979 golden KAT)
- `affinidi-messaging-didcomm`: 77 tests pass (all 16 JWS tests against the consolidated skeleton)
- mediator + mediator-setup: 594 tests pass (`--no-fail-fast`)
- `cargo fmt --check` clean; clippy clean on all targets, both default and `--features tsp`

## Note on the red CI checks

`publish dry-run` / `verify-publish` fail with E0432 on `verify_secp256k1` — that's the registry-isolated resolve not seeing the unpublished sibling crates (mediator → didcomm 0.15.5 → crypto 0.2.5), the expected-red pattern for multi-crate PRs introducing a cross-crate API. The release job publishes in dependency order; **merge as one PR, don't split it.**

One behavioral note for the changelog-conscious: the Ed25519 wrong-length-signature error string changed from `signature must be 64 bytes` to `EdDSA or Ed25519 signature must be 64 bytes` (error *variants* unchanged); called out in the didcomm changelog.

PR #646, opened 2026-07-23
fix: deliver TSP frames that arrive after the socket is open

stormer78:
A TSP send reported success and the frame was never delivered. It looked
intermittent — a probe would pass once, then fail for minutes — but it was
total: the outcome depended only on whether the frame reached the mediator
before or after the recipient's socket came up.

Reproduced with two local accounts on an embedded mediator, no VTA involved:
**0 of 10 frames delivered**. With this PR, 10 of 10.

## The two faults

**1. The mediator never marked a raw-TSP websocket live.** `Register` leaves a
client in `StreamingClientState::Registered` ("has a socket but has not enabled
live delivery"); only `Start` promotes it to `Live`. A DIDComm client gets there
by sending `messagepickup/3.0/live-delivery-change` — a raw-TSP socket carries
binary TSP frames and has no such message, so it stayed `Registered` for its
whole life. With the client not live, `store_message` skipped the streaming
publish, so the socket's re-drain arm never fired and flush-on-connect was the
only delivery it ever got. Everything arriving afterwards was stored and left in
the inbox.

Log signature: `TSP/bridged message stored for local recipient` with no
`Live streaming message to UUID` following it.

Raw-TSP delivery is push by contract (flush-on-connect + delete-on-send), not an
opt-in, so the socket now declares itself live when it registers.

**2. The SDK dropped packed frames that arrived between polls.** A TSP frame is
handed back packed (it can't be DIDComm-unpacked) and was delivered only if a
`Next` request happened to be outstanding at that instant, or a direct channel
was attached — otherwise it fell off the end of the function. The DIDComm branch
caches an unmatched message; the packed branch had nowhere to put one, since
`inbound_cache` is keyed by unpacked DIDComm message id. A polling consumer
always leaves a gap between one poll returning and the next being registered, so
this was a race the consumer could only lose.

Packed frames now go to a bounded `packed_cache` that the next `Next` drains,
and every arm hands the frame onward rather than dropping it.

## Also fixed

Both of these made the fault harder to diagnose than it should have been.

- **`Deregister` didn't check whose session it was.** `clients` is keyed by DID
  hash, so a socket that had already been replaced took its successor's channel
  down on the way out — and dropping that `tx` closes the live socket's command
  channel, so its handler exits with "streaming task unavailable". One DID
  handing off between transports (a DIDComm session closing as a raw-TSP socket
  opens) hits this routinely. Deregistration now names its session and is
  ignored if the slot has moved on.

- **`TspWebSocket::recv` discarded the close reason.** The mediator always says
  why it closed a socket ("replaced by a newer connection", "authentication
  token expired", "streaming task unavailable"); the client reported all of
  them, and a socket that simply went quiet, as the same bare `Ok(None)`. A
  close frame is now an `Err` carrying code and reason; a stream that ends
  without one stays `Ok(None)`. Signature unchanged.

## Tests

Every existing raw-TSP test queued its message *before* connecting, so they all
passed on flush-on-connect alone — which is how this shipped broken.

- `tsp_websocket_delivers_messages_that_arrive_after_connect` — connects first,
  against an empty inbox, then sends three frames.
- `tsp_frame_arriving_between_polls_is_not_dropped` — sleeps before polling, so
  the frame provably lands with no `Next` outstanding.

Both were checked to fail without their fix, with the symptom named in the
assertion. Full `affinidi-messaging-test-mediator`, mediator, sdk and
didcomm-service suites pass.

## Versions

`affinidi-messaging-mediator` 0.17.7, `affinidi-messaging-sdk` 0.18.62,
`affinidi-messaging-test-mediator` 0.2.40.

Behavioural change for downstream (R3.6): a raw-TSP socket now receives live
pushes, and `TspWebSocket::recv` returns `Err` where it previously returned
`Ok(None)` for a close frame. Consumers that treated `Ok(None)` as "socket gone"
should treat `Err` the same way.

## Not in this PR

Filed separately so this stays reviewable — see the linked issues: packed-cache
backpressure, the same session-id blindness in `Start`/`Stop`, delete-on-send
vs. R1.1, and the stale messaging CHANGELOG.

### DIFF
--- a/crates/messaging/affinidi-messaging-mediator/Cargo.toml
+++ b/crates/messaging/affinidi-messaging-mediator/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "affinidi-messaging-mediator"
-version = "0.17.6"
+version = "0.17.7"
 description = "Messaging Mediator service for Affinidi Messaging (DIDComm and TSP)"
 edition.workspace = true
 authors.workspace = true
--- a/crates/messaging/affinidi-messaging-mediator/src/handlers/websocket.rs
+++ b/crates/messaging/affinidi-messaging-mediator/src/handlers/websocket.rs
@@ -395,6 +395,41 @@ async fn handle_socket(
                     return;
                 }
             }
+
+            // Raw-TSP mode is live by construction — enable it here.
+            //
+            // `Register` alone leaves the client in `StreamingClientState::
+            // Registered` ("has a socket but has not enabled live delivery"),
+            // and only `Start` promotes it to `Live`. The DIDComm client gets
+            // there by sending `messagepickup/3.0/live-delivery-change`; a
+            // raw-TSP socket has no such message — it carries binary TSP frames
+            // and nothing else — so it stayed `Registered` forever. With the
+            // client not live, `store_message` skips the streaming publish
+            // entirely, so the `rx.recv()` re-drain arm below never fired and
+            // the ONLY delivery this socket ever got was flush-on-connect.
+            // Every frame that arrived after the socket came up was stored and
+            // left sitting in the inbox: the sender saw a successful send and
+            // the recipient heard nothing.
+            //
+            // Push delivery isn't an opt-in here the way it is for DIDComm —
+            // it is the raw-TSP contract (flush-on-connect + delete-on-send),
+            // so the socket declares itself live the moment it registers.
+            #[cfg(feature = "tsp")]
+            if tsp_mode {
+                let live = StreamingUpdate {
+                    did_hash: session.did_hash.clone(),
+                    state: StreamingUpdateState::Start,
+                };
+                match streaming.channel.send(live).await {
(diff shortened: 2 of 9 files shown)


stormer78:
Follow-ups filed so this PR stays focused:

- #647 — packed-frame cache sits outside the socket-read backpressure guard (deliberate; needs a deliberate policy decision either way)
- #648 — `Start`/`Stop` have the same session-id blindness this PR fixed for `Deregister`
- #649 — raw-TSP delete-on-send is at-most-once but documented as at-least-once (R1.1)
- #650 — `crates/messaging/CHANGELOG.md` is stale since mediator 0.15.5, so there's no current section to record this PR's behavioural change in

Downstream pickup (pin bump, mediator redeploy, and the `pnm health` probe correlation that let a stale queued reply pass as a pong): OpenVTC/verifiable-trust-infrastructure#749

PR #671, opened 2026-07-29
feat: BREAKING CHANGE: enforce authcrypt messages by default

ivan-branets-affinidi:


stormer78:
I pushed one commit on top of your rebase (`19df167`) — flagging it here so it doesn't get force-pushed away, since we were both working on this at the same time.

Two things the rebase onto `main` couldn't have caught, because they only became true once `main` moved:

**1. The version numbers now collide with published releases.** While this branch was open, `main` published `affinidi-messaging-didcomm` 0.15.6 and 0.15.7 (curve25519-dalek 5, then elliptic-curve 0.14) and `affinidi-messaging-test-mediator` 0.2.45 — the exact numbers this branch was claiming for different content. Since the release job skips versions that are already published, merging as-is would have quietly *not* published didcomm, then published `affinidi-messaging-sdk` 0.19.0 against a didcomm 0.15.7 that has no `VerifyKey` / `sign_multi`. Same shape as the drift incident we hit before.

Renumbered:
- `affinidi-messaging-didcomm` 0.15.7 → **0.15.8**, with the two colliding changelog entries collapsed into one
- `affinidi-messaging-mediator` 0.18.4 → **0.18.5** (its duplicated 0.18.1 entry moved up)
- `affinidi-messaging-test-mediator` 0.2.45 → **0.2.46**

`release guards` and the changelog guard both pass against current `main` now.

**2. The new multi-signature test code didn't build after the rebase.** `jws/sign.rs`, `jws/verify.rs` and `message/pack.rs` still called the pre-upgrade APIs, so `cargo test -p affinidi-messaging-didcomm` failed to compile:
- `rand_core::OsRng` → `rand_10::rng()` (rand_core 0.10 renamed `OsRng` to `SysRng` and made it fallible, so it no longer satisfies dalek's `CryptoRng` bound)
- `SigningKey::random(rng)` → `Generate::generate()`
- `to_encoded_point` → `to_sec1_point`

I also dropped the now-redundant `p256`/`k256` dev-dependencies — your rebase correctly put `ecdsa` on the runtime deps (`sign_multi` needs ES256/ES256K at runtime, not just in tests), which makes the test-only copies dead weight.

**3. Rollout section (R3.6).** The `affinidi-messaging-sdk` changelog now carries the piece that was still missing: the per-repo pin table (`vta-sdk` and `affinidi-trust-registry-rs` pin `affinidi-messaging-sdk = "0.18"` directly and will *not* pick up 0.19; `verifiable-trust-infrastructure` and `affinidi-webvh-service` get it via `affinidi-tdk = "0.8"`; `cierge` and `vti-message-bridge` only once `vta-sdk` moves). Worth calling out the sequencing hazard in there: once `affinidi-tdk` 0.8.5 pulls sdk 0.19 while a `^0.18` direct pin resolves to 0.18.x, any graph holding both links **two copies of the SDK** and `ATMConfig` / `UnpackMetadata` stop unifying — so `vta-sdk` and the trust registry need to move in the same cascade as the facade.

On senders: `vta-sdk`'s session transport already packs authcrypt, and the VTA has *required* authcrypt on `/auth/` and `/auth/refresh` since VTI #771, so the main path is fine. The one caveat is `vta-sdk::didcomm_light`, which is an anoncrypt packer whose output gets unpacked by `ATM::unpack` — no in-tree caller left, but the module is still `pub`, so any external caller will start getting `UnexpectedEnvelope` after this upgrade.

Verified locally on the pushed tree: `cargo fmt --check`, `cargo clippy --workspace --all-targets -D warnings`, and the full workspace test suite all clean. One unrelated pre-existing flake showed up under full-workspace parallelism — `mediator-setup`'s `config_writer::tests::test_all_secret_storage_refs` reads `std::env::current_dir()` while sibling tests in the same binary call `set_current_dir` process-wide. Passes in isolation, nothing to do with this PR, but probably worth its own fix.

Everything else in the review is addressed as far as I can tell — the layer-ordering fix with the ordered `CryptoLayer` stack, the `VerificationFailed` split for retry classification, the shared per-message resolution budget, and pinning the mediator message-id derivation with a real test rather than an assumption are all the right shape.

PR #674, opened 2026-08-01
feat!: move to elliptic-curve 0.14 and gate the legacy ssi-backed DID methods

stormer78:
Takes `p256`/`k256`/`p384`/`p521` 0.13 → 0.14, and puts `did:ethr` and `did:pkh`
behind features that are **off by default**.

## Why these are one change

Every workspace crate *could* move to the 0.14 family — but the graph kept **both**
generations anyway, because the `ssi-*` stack pins `p256`/`k256` 0.13. And `ssi-*`
was reachable only through `did:ethr` and `did:pkh`.

Those two were the **only** DID methods in the resolver SDK with no feature gate.
`did-jwk`, `did-webvh`, `did-cheqd`, `did-scid`, `did-ebsi` and `did_example` are
all already optional. An inconsistency in the gating was what held the old stack in
place.

```
resolver SDK, default build
  before:  elliptic-curve 0.13   p256 0.13   k256 0.13   ssi-jwk 0.3.2
  after:   elliptic-curve 0.14   p256 0.14   k256 0.14   (no ssi-* at all)
```

## ⚠️ BREAKING

**`did:ethr` and `did:pkh` no longer resolve unless `did-ethr` / `did-pkh` are
enabled.** Enable them to restore previous behaviour — at the cost of pulling
`ssi-*`, and with it a second elliptic-curve generation, back into your graph.
Resolution returns "method not supported" when off, exactly as the other gated
methods do. Verified both ways: with the features on, those resolvers still pass
their tests.

`ssi-dids-core` is now optional in both the resolver SDK and `did-scid`, reached
only via those features and `did-cheqd`. (`did-scid`'s only use of it was already
inside `#[cfg(feature = "did-cheqd")]`, which is not in default — so no code
change was needed there.)

**`affinidi-oid4vc-core`**: `Es256Signer::generate_with_rng` now bounds
`R: rand::CryptoRng` instead of the rand_core 0.6 traits, mirroring the
`EddsaSigner` change from the curve25519-dalek 5 move.

## 0.14 migration notes

- sec1 traits renamed: `ToEncodedPoint` → `ToSec1Point`, `FromEncodedPoint` → `FromSec1Point`
- `EncodedPoint` is now a *deprecated* alias for the generic `Sec1Point<C>`; it needs a per-module binding
- `SigningKey::random(rng)` / `SecretKey::random(rng)` are deprecated in favour of the `Generate` trait (`generate()` uses the system CSPRNG and panics on failure; `generate_from_rng(rng)` takes one)
- didcomm's `jwk` feature on p256/k256 is dropped — it doesn't exist in 0.14, and JWK handling goes through `affinidi-crypto`'s own `to_jwk`/`from_jwk`. It was dead weight already.

## Deliberately NOT included

**`affinidi-bbs` stays on elliptic-curve 0.13.** `bls12_381_plus` links both
generations by design — its `groups` feature enables `group_013` **and** `group`
0.14, with no way to pick one — so 0.13 is in the graph either way. Migrating BBS
to `hash2curve` 0.14 + sha2 0.11 would mean touching signature-critical crypto for
zero graph benefit. BBS is byte-identical to main.

Residual 0.13 therefore comes only from `bls12_381_plus` and `aws-sigv4` — both
external, both already at their latest version.

## Versioning

Patch bumps throughout, per [ADR 0003](docs/adr/0003-public-api-semver-policy.md)
point 3: `vta-sdk` and `didwebvh-rs` redirect these crates through
`[patch.crates-io]`, and a minor bump breaks the redirect. The behavioural break is
recorded in the changelog rather than the version number, per R3.6 — which is
exactly the case that ADR anticipates.

Seven crates: affinidi-crypto 0.2.7, affinidi-mdoc 0.2.5,
affinidi-did-resolver-cache-sdk 0.8.22, did-scid 0.1.13,
affinidi-messaging-didcomm 0.15.7, affinidi-messaging-mediator 0.18.3,
affinidi-oid4vc-core 0.1.7.

## Verification

- `cargo check --workspace --all-features --all-targets` clean
- **2958 tests pass, 0 failures** — identical to main
- **all 25 golden/KAT vectors pass**: `p256_sign_verify_roundtrip`,
  `secp256k1_es256k_golden` (deterministic RFC 6979), `ecdh_1pu_x25519_kek_golden`,
  `concat_kdf_es_golden`, `a256cbc_hs512_golden`, `aes256_kw_rfc3394_section_4_6`.
  Crypto behaviour is unchanged.
- clippy clean in **both** feature states — default, and with `did-ethr,did-pkh,did-jwk`
  enabled (CI builds `-D warnings`)
- `check-version-bumps.sh`, `check-changelogs.sh`, `check-workspace-duplicates.sh` green
- README support table updated — it listed `did:ethr`/`did:pkh` as unconditionally
  "Yes", which would have been wrong the moment this lands

### DIFF
--- a/crates/core/affinidi-crypto/CHANGELOG.md
+++ b/crates/core/affinidi-crypto/CHANGELOG.md
@@ -1,5 +1,24 @@
 # Affinidi Crypto Changelog
 
+## 2nd August 2026 (0.2.7)
+
+Moves this crate to the **elliptic-curve 0.14** family (`p256` / `k256` /
+`p384` / `p521` 0.13 -> 0.14), which brings rand_core 0.10, digest 0.11 and
+signature 3 with it.
+
+API changes handled: the sec1 traits were renamed (`ToEncodedPoint` ->
+`ToSec1Point`, `FromEncodedPoint` -> `FromSec1Point`), `EncodedPoint` became a
+generic alias for `Sec1Point<C>` so it needs a per-module binding, and the
+deprecated `SigningKey::random(rng)` is now `Generate::generate()` (system
+CSPRNG, panics on failure).
+
+Patch bump, per [ADR 0003](../../../docs/adr/0003-public-api-semver-policy.md) point 3: `vta-sdk` and `didwebvh-rs` redirect
+this crate through `[patch.crates-io]`, and a minor bump breaks the redirect.
+
+Crypto behaviour is unchanged — the golden/KAT vectors (`p256_sign_verify_roundtrip`,
+`secp256k1_es256k_golden` RFC 6979, `ecdh_1pu_x25519_kek_golden`,
+`concat_kdf_es_golden`, `a256cbc_hs512_golden`) all still pass.
+
 ## 31st July 2026 (0.2.6)
 
 **`PrivateKeyAgreement::X25519` now holds `[u8; 32]` instead of an
--- a/crates/core/affinidi-crypto/Cargo.toml
+++ b/crates/core/affinidi-crypto/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "affinidi-crypto"
-version = "0.2.6"
+version = "0.2.7"
 description = "Cryptographic primitives and JWK types for Affinidi TDK"
 edition.workspace = true
 authors.workspace = true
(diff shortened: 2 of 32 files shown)


stormer78:
### `publish dry-run` is red, and that is expected here

```
failed to select a version for the requirement `affinidi-did-resolver-cache-sdk = "^0.8.22"`
```

`affinidi-did-resolver-cache-server` now requires the SDK at **0.8.22**, because
that is the first version with the `did-ethr` / `did-pkh` features it needs to
opt back into. 0.8.22 is published by *this* PR, so a registry-isolated resolve
cannot see it yet.

The requirement genuinely has to move: leaving it at `0.8.19` would let cargo
resolve a published SDK that has no `did-ethr` feature, and the server would
fail to build for anyone outside this workspace.

The release job publishes in dependency order — SDK 0.8.22 first, then the
server against it — so this resolves on merge. Same class of expected-red as
affinidi/affinidi-tdk-rs#675 had for `MediatorBuilder::listener`. **Please don't
split the PR to work around it**; the split halves would hit the same red.

Everything else is green.

PR #675, opened 2026-08-01
fix: hand the test mediator's listener over instead of re-binding its port

stormer78:
## The race

`TestMediator` bound `127.0.0.1:0` to learn the ephemeral port, **dropped** the
listener, then passed the address to `MediatorBuilder::listen_addr` so the server
could bind that same port again. Between the drop and the re-bind the port is
free, and another process can take it.

It did — on #673's Coverage job:

```
env spawn: Mediator(InternalError(17, "NA",
  "Failed to bind 127.0.0.1:38221: Address already in use (os error 98)"))
```

`add_admin_rejects_mismatched_identity` lost that race. It surfaced under
`cargo llvm-cov` rather than the plain test run because the slower,
differently-timed execution widens the window — which is why Coverage failed
while Test Suite passed on the same commit.

The helper's own comment had already called this out and prescribed the fix:

> *"There is a tiny TOCTOU window between drop and re-bind during which another
> process could claim the port… If this proves flaky in CI, extend
> `MediatorBuilder` to accept a pre-bound `TcpListener` and remove the rebind
> step."*

## Why the pre-bind can't just be deleted

The harness needs the bound port **before** the server starts — it mints the
mediator's `did:peer` whose service endpoint embeds the URL. So the listener is
handed over rather than the bind removed.

## The fix

- **`MediatorBuilder::listener(TcpListener)`** — additive. Takes a listener the
  caller already bound and reads the address back off it. `listen_addr` is
  unchanged and still the right choice when you don't need the port in advance.
- `StartOpts` carries the listener through; `serve_internal` uses it when present
  and otherwise binds exactly as before — **the TOML/binary path is untouched**.
- `bind_ephemeral_listener` now returns `(TcpListener, SocketAddr)`, and the
  harness passes the listener to the builder. **The port is never released**, so
  the window is gone.

## Regression guard

New test `ephemeral_listener_keeps_the_port_bound` asserts the port is *not*
re-bindable while we hold the listener:

```rust
let (listener, addr) = bind_ephemeral_listener(None).expect("bind");
assert!(TcpListener::bind(addr).is_err(), "…the listener is not holding it");
```

If that ever passes a second bind, the race is back.

## Verification

- `cargo check --workspace --all-features --all-targets` clean
- **2959 tests pass, 0 failures** (2958 as on main, plus the new guard)
- default-feature suite green as well, not just `--all-features`
- lifecycle suite — the one that flaked — run **5× consecutively, 22 passed each time**
- clippy clean on both feature sets (default redis-backend, and
  `--no-default-features --features didcomm,memory-backend`)
- `cargo fmt --all --check` clean; version-bump and changelog guards green

## Note for whoever merges second

This and #674 both bump `affinidi-messaging-mediator` 0.18.2 → 0.18.3, so they
will conflict on the version line and changelog. Whichever lands second needs a
rebase and a re-bump to 0.18.4 — a one-line fix.

### DIFF
--- a/crates/messaging/affinidi-messaging-mediator/CHANGELOG.md
+++ b/crates/messaging/affinidi-messaging-mediator/CHANGELOG.md
@@ -1,5 +1,20 @@
 # Affinidi Messaging Mediator
 
+## 2nd August 2026 (0.18.3)
+
+Adds **`MediatorBuilder::listener`**, which hands the server a listener the
+caller has already bound instead of making it bind for itself.
+
+A caller that needs the bound port *before* startup — to mint a DID whose
+service endpoint embeds the URL, say — previously had to bind, read the port,
+drop the listener, and pass the address for the server to re-bind. That leaves
+a window between the drop and the re-bind in which another process can take the
+port. Handing the listener across removes the window: the port is never
+released.
+
+Additive. `listen_addr` is unchanged and still the right choice when you do not
+need the port in advance; the TOML/binary path is untouched.
+
 ## 1st August 2026 (0.18.2)
 
 Drops the **legacy rustls 0.21 stack** from the AWS SDK dependency path,
--- a/crates/messaging/affinidi-messaging-mediator/Cargo.toml
+++ b/crates/messaging/affinidi-messaging-mediator/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "affinidi-messaging-mediator"
-version = "0.18.2"
+version = "0.18.3"
 description = "Messaging Mediator service for Affinidi Messaging (DIDComm and TSP)"
 edition.workspace = true
 authors.workspace = true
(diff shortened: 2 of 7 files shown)


stormer78:
### `publish dry-run` / `verify-publish` are red, and that is expected here

```
error[E0599]: no method named `listener` found for struct `MediatorBuilder`
   822 |             .listener(listener)
```

Both jobs resolve each crate against **crates.io in isolation**, so
`affinidi-messaging-test-mediator` is built against the *published*
`affinidi-messaging-mediator` 0.18.2 — which does not have
`MediatorBuilder::listener` yet, because this PR is what adds it.

This is the known expected-red case for a PR that adds a cross-crate API: the
release job publishes in dependency order (mediator 0.18.3 first, then
test-mediator 0.2.45 against it), so it resolves on merge. **Please don't split
the PR to work around it** — splitting would ship a `listener` method with no
caller, and the second half would hit the same red anyway.

Everything else is green: 41 pass, and the two reds are exclusively this
resolution artefact.

PR #689, opened 2026-08-08
feat: add DIDComm v1 forward ingress to the mediator

stormer78:
Adds DIDComm v1 (Aries RFC 0019) **forward ingress** to the mediator, behind a
new `didcomm-v1` feature. Follows the `tsp` precedent: additive, compile-time,
off by default. A mediator built or configured without it behaves exactly as
before.

This is stage 1 of the two agreed. Stage 2 brings the Aries
`coordinate-mediation/1.0` and `messagepickup` protocols — see *Not yet usable
by a wallet* below.

## What it does

```text
anoncrypt(to = mediator routing verkey)
  └── { "@type": ".../routing/1.0/forward",
        "to":  "<recipient base58 verkey>",
        "msg": { …inner envelope, opaque to the mediator… } }
```

The mediator opens only the outer layer, resolves the `to` verkey to a local
account, applies that account's ACLs, and stores the inner envelope verbatim for
pickup. The inner bytes are never parsed — they are sealed end-to-end to the
recipient, exactly as the TSP relay path treats an opaque inner.

### 1. The sniff needs two discriminators

A v2 JWE and a v1 envelope share `protected` / `iv` / `ciphertext` / `tag`, so
shape alone cannot separate them. v1 is identified by having **no top-level
`recipients`** (it nests them inside the protected header) **and** declaring
`"typ": "JWM/1.0"`. Requiring both means an unknown envelope falls through to the
v2 path and gets v2's error, rather than being misreported as broken v1.

### 2. Anonymous admission is v1-scoped, deliberately

An RFC 0019 forward is anon-packed *by design*, so there is no authenticated
sender to apply a sender-side ACL to — the recipient's access list carries the
weight. That is a real departure from this mediator's posture, so it is opt-in
per deployment via `[didcomm_v1] allow_unauthenticated_forwards` (default
`false`), and it is what stock Credo wallets need, since they cannot complete
the v2 challenge.

Turning it on must not become a general anonymous-inbound licence. So
`anonymous_session_for` yields a **v1-scoped** session when admitted only by that
flag:

- the handler refuses any non-v1 body on it (`ANON_V1_SESSION_ID`);
- as defence in depth the session carries `SEND_MESSAGES` **without**
  `SEND_FORWARDED`, so it could not drive a v2 relay forward even if the body
  check were bypassed;
- inter-mediator relay admission keeps precedence and its existing capabilities;
- a *presented but invalid* credential is still never downgraded to anonymous.

Four tests cover exactly these.

### 3. A verkey binds to at most one DID

Otherwise any account could claim another's routing key and capture its inbound
v1 traffic — a delivery-hijack primitive, not a tidiness issue. Redis enforces it
with `SETNX`; the in-process backends do it under their write lock. The
conformance suite asserts the second claim is refused **and** that the failed
claim leaves the original owner intact.

## The store layer

`MediatorStore` gains `v1_routing_key_bind` / `_lookup` / `_unbind` /
`v1_routing_keys_for`, plus a `supports_v1_routing_keys()` probe. All have
defaults so third-party backends keep compiling, and the probe defaults to
`false` — **startup refuses to run** with v1 enabled on a backend that returns
it, rather than accepting traffic it could not route.

The index stores the **DID**, not its hash. I wrote it hash-keyed first, then
found `WrapperType::Envelope` takes a DID and hashes it internally: a hash cannot
be reversed, so ingress would have had nothing to hand the store path. The
*reverse* index stays hash-keyed because that is what `account_remove` works
from — and it does purge, or a removed account's verkeys keep resolving to a
mailbox that no longer exists.

Implemented for all three backends (memory, fjall, redis) with conformance
coverage.

## No new key material

The mediator's v1 routing verkey is derived from the Ed25519 authentication key
it already holds — v1 does key agreement on that key's Montgomery form. Same
trick `tsp_identity` uses: one set of operating secrets, three protocols. Pinned
in a test against the verkey a real Credo agent produced for the same seed.

It is logged at startup, and the identity is derived eagerly so a missing key
fails at boot rather than on the first inbound forward. v1 predates DID
documents as a discovery mechanism, so operators have to hand this to clients out
of band and have nowhere else to read it from.

## mediator-setup

`didcomm-v1` becomes a recipe protocol and writes a `[didcomm_v1]` section.
Selecting it turns `didcomm` on too, matching the Cargo feature — otherwise a
recipe listing only `didcomm-v1` produces a build whose features imply v2.1 while
the wizard believes it is off. The section is written only when enabled, and a
re-run that disables v1 strips it rather than leaving a stale `enabled = true`.

## Not yet usable by a wallet

Deliberately, per the staged plan. There is no client-facing way to register a
routing key until `coordinate-mediation/1.0` lands, so bindings can only be
created through the store API and a stock Credo wallet cannot yet use this
mediator end-to-end. Stage 2 adds registration and pickup.

## Incidental

Corrects the fjall keyspace count in three places (14 → 15) now that the routing
index adds one.

## Verification

- `cargo test -p affinidi-messaging-mediator` — 254 pass on
  `didcomm-v1,memory-backend,fjall-backend`; 182 lib tests on the default build
- `cargo test -p affinidi-messaging-mediator-setup` — 406 pass
- Store conformance green on memory + fjall (Redis paths are `REDIS_URL`-gated
  as usual)
- `cargo clippy --workspace --all-targets` clean; mediator clean on default,
  `didcomm-v1`, and `didcomm-v1,tsp`
- `check-version-bumps.sh`, `check-changelogs.sh`,
  `check-workspace-duplicates.sh` — all pass

### DIFF
--- a/crates/messaging/affinidi-messaging-mediator/CHANGELOG.md
+++ b/crates/messaging/affinidi-messaging-mediator/CHANGELOG.md
@@ -1,5 +1,58 @@
 # Affinidi Messaging Mediator
 
+## 8th August 2026 (0.18.8)
+
+**DIDComm v1 (Aries RFC 0019) forward ingress — new `didcomm-v1` feature.**
+
+Off by default; a mediator built or configured without it behaves exactly as
+before. Enabling it lets the mediator act as an Aries mediator for inbound
+forwards: it opens the outer anoncrypt layer with its own routing key, resolves
+the forward's `to` **verkey** through the new routing-key index, applies the
+recipient's ACLs, and stores the inner envelope verbatim for pickup. The inner
+bytes are never parsed — they are sealed end-to-end to the recipient.
+
+Requires the `didcomm` feature (v1 is additive over v2.1, not a replacement) and
+a storage backend implementing the v1 routing-key index. Startup **refuses to
+run** with `[didcomm_v1] enabled = true` on a backend that does not, rather than
+accepting traffic it could not route.
+
+### New `[didcomm_v1]` configuration
+
+```toml
+[didcomm_v1]
+enabled = true
+# Aries mediators accept anon-packed forwards from anyone; this mediator
+# normally requires an authenticated session. Off by default.
+allow_unauthenticated_forwards = false
+```
+
+`allow_unauthenticated_forwards` is a **deliberate, per-deployment loosening**
+of this mediator's posture, and is required for stock Credo wallets, which
+cannot complete the v2 authentication challenge. An RFC 0019 forward is
+anon-packed by design, so there is no authenticated sender to apply a
+sender-side ACL to and the recipient's access list carries the weight.
+
+Turning it on does **not** open anonymous DIDComm v2 inbound. A session admitted
+only by this flag is v1-scoped: the handler refuses any non-v1 body on it, and
+the session carries `SEND_MESSAGES` without `SEND_FORWARDED`, so it cannot drive
+a v2 relay forward even if that check were bypassed. Inter-mediator relay
+admission keeps precedence and its existing capabilities, and a *presented but
+invalid* credential is still never downgraded to anonymous.
+
+### Routing key
+
+The mediator's v1 routing verkey is derived from the Ed25519 authentication key
+it already holds — no new key material, the same trick `tsp_identity` uses. It
+is logged at startup because v1 predates DID documents as a discovery
+mechanism, so operators must hand it to clients out of band.
+
+### Not yet included
+
+The Aries `coordinate-mediation/1.0` and `messagepickup` protocols. Until those
+land there is no client-facing way to register a routing key, so a stock Credo
+wallet cannot yet use this mediator end-to-end — bindings can only be created
+through the store API.
+
 ## 8th August 2026 (0.18.7)
 
 **Envelope-wrapping observability — warn on non-conformant layering.**
--- a/crates/messaging/affinidi-messaging-mediator/Cargo.toml
+++ b/crates/messaging/affinidi-messaging-mediator/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "affinidi-messaging-mediator"
-version = "0.18.7"
+version = "0.18.8"
 description = "Messaging Mediator service for Affinidi Messaging (DIDComm and TSP)"
 edition.workspace = true
 authors.workspace = true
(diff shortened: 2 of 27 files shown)


stormer78:
### Branch synced with main, and a note on the red `publish dry-run`

Merged `main` (post-#688, so `affinidi-messaging-didcomm-v1` is 0.1.1 here). No
conflicts; guards, build and the 182 lib tests still green.

**`publish dry-run` is red, and it is the expected multi-crate case — please
don't "fix" it by loosening the pin.**

Only one of the three crates fails, and only on an unpublished sibling:

```
--- cargo publish -p affinidi-messaging-mediator --dry-run ---
error: failed to prepare local package for uploading
  failed to select a version for the requirement `affinidi-messaging-mediator-common = "^0.15.34"`
  candidate versions found which didn't match: 0.15.33, 0.15.32, 0.15.31, ...
  FAIL affinidi-messaging-mediator

--- cargo publish -p affinidi-messaging-mediator-common --dry-run ---
   Packaged 74 files ... Uploading affinidi-messaging-mediator-common v0.15.34   ✅

--- cargo publish -p affinidi-messaging-mediator-config --dry-run ---
   Packaged 15 files ... Uploading affinidi-messaging-mediator-config v0.2.1     ✅
```

The dry-run resolves each crate against the registry in isolation, so it cannot
see that `mediator-common` 0.15.34 is being published by this same PR. The
release job publishes in dependency order (common → config → mediator), so it
resolves on merge. Same shape as the previously documented multi-crate case.

**The `^0.15.34` pin is load-bearing.** The mediator calls
`supports_v1_routing_keys()` and `v1_routing_key_lookup()`, which only exist from
0.15.34. Relaxing it to `"0.15"` would make the dry-run green and ship a crate
that fails to compile for anyone whose lockfile holds 0.15.33 — which is exactly
the bug #688 just fixed one layer up. Leaving it tight moves that failure from a
confusing `E0599` deep in a dependency to a resolution error naming the real
constraint.

stormer78:
### CI settled: 41 pass, 5 skipped, 2 fail — both the same expected cause

Correcting the scope of my previous comment: `verify-publish` fails too, not just
`publish dry-run`, and for the identical reason. It packages the whole workspace
and stops at the first failure, which is the mediator:

```
error: failed to prepare local package for uploading
  failed to select a version for the requirement `affinidi-messaging-mediator-common = "^0.15.34"`
  candidate versions found which didn't match: 0.15.33, 0.15.32, 0.15.31, ...
```

So both red checks are the same registry-isolation artefact, and every other
check — including `release guards`, clippy, fmt, the full test suite, all storage
and secret-backend matrices, and `test-mediator (memory/fjall/tsp)` — is green.

The reasoning in the previous comment stands: the `^0.15.34` pin is load-bearing
and both checks go green on merge once the release job publishes
`mediator-common` ahead of `mediator`.

PR #690, opened 2026-08-08
feat: add v1 coordinate-mediation and message-pickup to the mediator

stormer78:
Stage 2 of the mediator's DIDComm v1 support, completing what #689 started: a v1
wallet can now register its routing keys and collect what arrives, so the
`didcomm-v1` feature is usable end to end rather than plumbing.

Adds **coordinate-mediation 1.0** (RFC 0211) and **message-pickup 2.0**
(RFC 0685) — message types in the v1 crate, handlers in the mediator.

## The whole flow, tested against a live mediator

`tests/didcomm_v1_mediation.rs` walks the exact sequence an Aries wallet
performs, over HTTP, nothing stubbed:

```
mediate-request   →  mediate-grant            learn the routing key
keylist-update    →  keylist-update-response  register a recipient key
                     (third party forwards)   a message arrives
status-request    →  status                   one queued
delivery-request  →  delivery                 the message, still queued
messages-received →  status                   acknowledged and dropped
```

Each reply is opened by the client, which is what proves the mediator packed it
correctly — the response path is covered as well as the request path. The
delivered envelope is opened by its real recipient, showing the end-to-end
sender survives the round trip untouched.

### 1. Replies go back as the raw HTTP body

That is Aries' return-route: a one-way transport carrying a synchronous
response. Credo's `isValidJweStructure` requires only `protected` / `iv` /
`ciphertext` / `tag`, so a v1 envelope is accepted there unchanged — wrapping it
in the mediator's JSON success envelope would hand the client a reply it cannot
parse. `/inbound` now returns a `Response` so the v1 branch can answer directly.

Replies are **authcrypt**, not anoncrypt: a forgeable `mediate-grant` would let
an attacker nominate its own routing keys and capture the client's inbound
traffic.

### 2. A third dual-spelling trap

`recipient_key` and `routing_keys` accept bare base58 **or** `did:key` for the
same key. RFC 0211 specifies `did:key`, base58 predates it and is still widely
sent, and Credo normalises on receipt while picking its outbound form from
`useDidKeyInProtocols`. Same shape as the two message-type document URIs found in
stage 1, one field down. `Verkey::parse` handles it and comparisons happen on the
decoded key; there is a test that a key registered as `did:key` is removable as
base58.

### 3. `mediate-request` is denied when the account cannot receive

The e2e test caught this. An auto-created account inherits `global_acl_default`,
which by default does not grant `RECEIVE_MESSAGES` — so the client was told
mediation was granted, published its routing keys, and then every forward was
refused with 403. A silent black hole with no way for the client to detect it.

`mediate-request` now checks and answers `mediate-deny` instead, with a test for
each branch. **Operators serving v1 wallets need a `global_acl_default` that
grants receive**; the changelog says so.

## Identity and admission

A client's account is the `did:key` derived from its authenticated verkey —
deterministic, no extra state. Unlike forwards, these messages are authenticated
by construction: opening the envelope proves which verkey sent it, and a v1
plaintext has no `from` header to lie in. An anonymous mediation request is
refused.

Provisioning mirrors `handlers::authenticate::response` rather than adding a
second policy: unknown accounts are created with `global_acl_default` unless
`mediator_acl_mode` is `explicit_allow`. One admission gate, two entry points.

## Handler behaviour

- `keylist-update` reports a result per key rather than failing the batch, as
  RFC 0211 expects. Claiming or removing another account's key is refused as
  `client_error` — the same answer as any other rejected entry, so it cannot be
  used to probe which keys others hold.
- `delivery-request` fetches with `DoNotDelete`: pickup acknowledges separately,
  and deleting on delivery loses any message whose delivery was lost in flight.
  `messages-received` deletes under `Owner` authority, so a client cannot delete
  another's messages by guessing ids.
- An empty queue answers `status`, not an empty `delivery`, which would read as
  a delivery that dropped everything.
- Delivery limits capped, acknowledgement lists bounded — authenticated, but the
  numbers are still caller-chosen.
- `live-delivery-change` answers truthfully with `live_delivery: false` rather
  than accepting and never pushing.

## A test that passed for the wrong reason

`forward_to_an_unregistered_key_is_refused` asserted 404 and got it — from a
mistyped URL, not from the mediator refusing anything. It now probes the route
first, so the assertion cannot pass unless `/inbound` exists.

## Scope

Pickup **2.0** only: Credo's default strategy is `PickUpV2` and 1.0
(`batch-pickup`) is legacy, worth adding if a target wallet needs it. Live
delivery is answered truthfully rather than implemented — it needs a persistent
connection per client, which is a separate piece of work.

## Verification

- `cargo test -p affinidi-messaging-test-mediator --features didcomm-v1` — 82 pass
- `cargo test -p affinidi-messaging-didcomm-v1` — 122 pass
- `cargo test -p affinidi-messaging-mediator --features didcomm-v1` — 185 lib pass
- `cargo clippy --workspace --all-targets` clean; mediator clean on default,
  `didcomm-v1`, and `didcomm-v1,tsp`
- `check-version-bumps.sh`, `check-changelogs.sh`,
  `check-workspace-duplicates.sh` — all pass

`publish dry-run` / `verify-publish` will be red on the same unpublished-sibling
grounds as #689: `affinidi-messaging-mediator` requires
`affinidi-messaging-didcomm-v1 = "0.2"`, which this PR publishes. The release job
publishes in dependency order.

### DIFF
--- a/crates/messaging/affinidi-messaging-didcomm-v1/CHANGELOG.md
+++ b/crates/messaging/affinidi-messaging-didcomm-v1/CHANGELOG.md
@@ -1,5 +1,40 @@
 # affinidi-messaging-didcomm-v1 changelog
 
+## 0.2.0 — 8th August 2026
+
+Adds the two Aries protocols a wallet needs to use a mediator.
+
+### Added
+
+- **`protocols::coordinate_mediation`** (RFC 0211) — `mediate-request` /
+  `-grant` / `-deny`, `keylist-update` and its response, `keylist-query` /
+  `keylist`.
+- **`protocols::message_pickup`** (RFC 0685, 2.0) — `status-request` /
+  `status`, `delivery-request` / `delivery`, `messages-received`,
+  `live-delivery-change`, plus `~transport.return_route` helpers.
+- **`Verkey::parse` / `from_did_key` / `to_did_key`** — a key field written as
+  bare base58 or as `did:key` names the same key, and both spellings are on the
+  wire.
+
+### Wire-format notes
+
+- **A third dual-spelling trap.** `recipient_key` and `routing_keys` accept
+  base58 *or* `did:key`: RFC 0211 specifies the latter, base58 predates it and
+  is still widely sent, and Credo normalises on receipt while choosing its
+  outbound form from `useDidKeyInProtocols`. Comparing the strings drops half
+  the ecosystem — the same shape as the two message-type document URIs, one
+  field down. Parse, then compare `Verkey`s. This crate emits base58.
+- **Delivery is not deletion.** A `delivery` removes nothing; only
+  `messages-received` may. A mediator that deletes on delivery loses any
+  message whose delivery was lost in flight.
+- **`~transport: { return_route: "all" }`** is how these requests expect their
+  reply in the HTTP response body. Credo's `isValidJweStructure` requires only
+  `protected` / `iv` / `ciphertext` / `tag`, so a v1 envelope is accepted there
+  unchanged.
+
+Scoped to pickup 2.0: Credo's default strategy is PickUpV2 and 1.0
+(`batch-pickup`) is legacy.
+
 ## 0.1.1 — 8th August 2026
 
 Manifest fix. No code change.
--- a/crates/messaging/affinidi-messaging-didcomm-v1/Cargo.toml
+++ b/crates/messaging/affinidi-messaging-didcomm-v1/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "affinidi-messaging-didcomm-v1"
 description = "DIDComm v1 (Aries RFC 0019) messaging implementation for the Affinidi TDK"
-version = "0.1.1"
+version = "0.2.0"
 edition.workspace = true
 authors.workspace = true
 readme = "README.md"
--- a/crates/messaging/affinidi-messaging-didcomm-v1/src/identity.rs
+++ b/crates/messaging/affinidi-messaging-didcomm-v1/src/identity.rs
@@ -146,6 +146,38 @@ impl Verkey {
         Ok(Self(bytes))
     }
 
+    /// Decode a verkey written **either** as bare base58btc **or** as a
+    /// `did:key:z6Mk…`.
+    ///
+    /// Both spellings appear on the wire for the same key, and which one a peer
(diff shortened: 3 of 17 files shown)


stormer78:
### CI settled: 41 pass, 2 fail — both the expected unpublished-sibling case

Fixed since opening: a stolen doc comment on `TestMediatorBuilder::acl_mode`
(the crate denies `missing_docs`, which only fires under CI's
`RUSTFLAGS=-D warnings`), a clippy `?`-operator lint in the e2e test, and a
missing `affinidi-messaging-test-mediator` version bump. Re-verified the whole
matrix under CI's exact flags — workspace default, `didcomm-v1`,
`didcomm-v1,tsp`, and the v1 crate — all clean.

**`checks-storage / fjall` was an infrastructure flake, not code.** It failed in
12s at `Install native deps for aws_lc_rs + keyring`, before the Rust toolchain
was installed, with every cargo step `skipped`. The four sibling jobs that do
exercise fjall — `test-mediator (fjall)`, `fjall + aws`, `fjall + vault`,
`all-backends compile` — all passed in the same run, and the previous run of the
same workflow on this branch was green. Re-run, now green.

**The two remaining failures are the documented multi-crate case.** Only the
crates depending on the version this PR publishes fail:

```
--- cargo publish -p affinidi-messaging-didcomm-v1 --dry-run ---
   Uploading affinidi-messaging-didcomm-v1 v0.2.0                              ✅

--- cargo publish -p affinidi-messaging-mediator --dry-run ---
  failed to select a version for the requirement `affinidi-messaging-didcomm-v1 = "^0.2"`
  candidate versions found which didn't match: 0.1.1, 0.1.0
  FAIL affinidi-messaging-mediator

--- cargo publish -p affinidi-messaging-test-mediator --dry-run ---
  (same)
```

The dry-run resolves each crate against the registry in isolation, so it cannot
see that `didcomm-v1` 0.2.0 is being published by this same PR. The release job
publishes in dependency order (didcomm-v1 → mediator → test-mediator), so both
checks go green on merge. Same shape as #689.

The `^0.2` requirement is load-bearing: the mediator uses
`protocols::coordinate_mediation` and `protocols::message_pickup`, which only
exist from 0.2.0. Relaxing it to `"0.1"` would make CI green and ship a crate
that fails to compile against a 0.1.x lockfile.

PR #714, opened 2026-08-16
chore(deps): track trust-tasks-rs 0.9.0 across the messaging stack

stormer78:
## What

Moves the six messaging crates from `trust-tasks-rs` 0.6.1 to **0.9.0**.

| crate | version |
|---|---|
| `affinidi-messaging-sdk` | 0.19.7 → 0.19.8 |
| `affinidi-messaging-mediator` | 0.18.17 → 0.18.18 |
| `affinidi-messaging-didcomm-service` | 0.3.25 → 0.3.26 |
| `affinidi-messaging-test-mediator` | 0.2.50 → 0.2.51 |
| `affinidi-messaging-text-client` | 0.12.5 → 0.12.6 |
| `affinidi-messaging-helpers` | 0.13.3 → 0.13.4 |

## What changed upstream, and what reaches us

Three breaking releases sit between 0.6.1 and 0.9.0. **One** reaches the source.

- **0.7.0** — `StandardCode` became `#[non_exhaustive]`. Nothing in this repo matches on it, so no arm had to be added. It is also the last time a new framework error code will break a downstream `match`.
- **0.8.0** — added `StandardCode::Cancelled` and `trust-task-control/0.1`; emitted error documents moved to `trust-task-error/0.5`. Additive in Rust because of the attribute above.
- **0.9.0** — `consume_inbound` gained a required `PayloadPolicy` argument (SPEC.md §7.2 item 2), and `ValidatedPayload::SCHEMA_JSON` was replaced by `Payload::PAYLOAD_SCHEMA`. **One** call site in the repo (the mediator's `ping` consume path); `SCHEMA_JSON` had no users.

## The one behavioural decision, and it is a deliberate no-op

The `ping` path passes `PayloadPolicy::<NoValidator>::AcceptUnvalidated` — the behaviour the call has always had, now stated rather than implied. `ping::v0_1::Payload` is a generated struct with `deny_unknown_fields` and validating newtypes, so deserializing into it already enforced the required members, their types, and the pattern/length constraints. What `Validate` would add is the residue a Rust type cannot express (`minProperties`, `minItems` on an optional array, conditional subschemas) — and the `ping` spec declares none of it.

**No document is accepted or refused differently as a result of this PR.** Moving to `Validate` can begin refusing documents a peer sends today, so it belongs in its own release with its own rollout, never folded into a dependency bump.

## Why patches, not minors

`trust-tasks-rs` is a public dependency of the SDK's `trust_tasks()` surface, so on the face of it this is breaking. It ships as a patch anyway, following the convention 0.19.2 and 0.19.7 set and for the same reason: every consumer requires `affinidi-messaging-sdk` `^0.19` and resolves it from the registry, so a `0.20` would pull a **second registry copy of the SDK** into their graphs — the exact defect this bump exists to remove, relocated one crate over.

The cost is unchanged and is stated in each changelog: this is **source-breaking for a consumer still on `trust-tasks-rs` 0.6.x** despite the patch version. Those consumers move their own pin in the same change.

## The other `trust-tasks-rs` in the lockfile is untouched

The lock still carries `trust-tasks-rs` 0.2.57, reached through `vta-sdk` 0.21.6 behind the mediator's `vta-sdk = ">=0.21.0, <0.23.0"` ceiling. It stays **inert**: no file in the mediator uses both `vta_sdk` and `trust_tasks_rs`, so no trust-tasks type crosses that boundary and the two copies never have to unify.

Collapsing it needs a `vta-sdk` release built on 0.9, which does not exist yet — that is a later, separate PR, deliberately not bundled here.

## Verification

- `cargo check --workspace --all-targets` — clean
- `cargo check --workspace --all-features --all-targets` — clean
- `cargo test --workspace` — all suites pass, 0 failures
- `cargo fmt --all --check`, `cargo clippy --workspace --all-targets` — clean
- All four release guards run locally and pass: version-bump, changelog, workspace-duplicates, toolchain-sync (plus test-clock isolation)

### DIFF
--- a/crates/messaging/affinidi-messaging-didcomm-service/CHANGELOG.md
+++ b/crates/messaging/affinidi-messaging-didcomm-service/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Changelog
 
+## [0.3.26] - 2026-08-17
+
+### Changed
+
+- **Track `trust-tasks-rs` 0.9.0** and `affinidi-messaging-sdk` 0.19.8, up from
+  0.6.1 / 0.19.7. Version requirements only — the single use here is
+  `specs::messaging::account`, whose `v0_1` types are shape-identical across the
+  move. See the SDK's 0.19.8 entry for what the three intervening breaking
+  releases changed and why this ships as a patch.
+
 ## [0.3.25] - 2026-08-14
 
 ### Changed
--- a/crates/messaging/affinidi-messaging-didcomm-service/Cargo.toml
+++ b/crates/messaging/affinidi-messaging-didcomm-service/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "affinidi-messaging-didcomm-service"
-version = "0.3.25"
+version = "0.3.26"
 description = "Shared DIDComm service framework for always-online DIDComm client. Manages mediator connection, message lifecycle, and handler dispatch."
 edition.workspace = true
 authors.workspace = true
@@ -21,10 +21,10 @@ tsp = ["affinidi-messaging-sdk/tsp"]
 
 [dependencies]
 affinidi-tdk-common = "0.6"
-affinidi-messaging-sdk = "0.19.7"
+affinidi-messaging-sdk = "0.19.8"
 affinidi-messaging-didcomm = "0.15"
 ## Trust Tasks payload types (the atm.trust_tasks() surface accepts these).
-trust-tasks-rs = "0.6.1"
+trust-tasks-rs = "0.9"
 affinidi-secrets-resolver = "0.5"
 
 async-trait = "0.1"
--- a/crates/messaging/affinidi-messaging-helpers/Cargo.toml
+++ b/crates/messaging/affinidi-messaging-helpers/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "affinidi-messaging-helpers"
-version = "0.13.3"
+version = "0.13.4"
 description = "CLI tools for Affinidi Messaging (mediator administration)"
 edition.workspace = true
 authors.workspace = true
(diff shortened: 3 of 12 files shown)


stormer78:
### Note on the one red check — `verify-publish / Verify Publishable`

This failure is expected for this shape of change and matches the merged precedent exactly.

`cargo package` resolves dependencies from **crates.io**, not the workspace. Three crates pin the SDK by exact patch:

```
error: failed to prepare local package for uploading
  failed to select a version for the requirement `affinidi-messaging-sdk = "^0.19.8"`
```

`affinidi-messaging-sdk 0.19.8` does not exist on the registry yet — it is published *by merging this PR*. So `affinidi-messaging-{didcomm-service,test-mediator,mediator}` cannot be packaged until after the SDK lands, and the job cannot pass before merge.

**Same job, same three crates, same error on #709** (the 0.6.1 bump, merged 2026-08-14):

```
failed to select a version for the requirement `affinidi-messaging-sdk = "^0.19.7"`
```

That one was merged with this check red, and the release pipeline published the crates in dependency order afterwards.

**Loosening the pin to `^0.19` is not the fix.** The exact pairing is load-bearing and deliberate — `affinidi-messaging-mediator`'s own changelog records why 0.18.15 adopted it: `trust-tasks-rs` is a public dependency of the SDK's `trust_tasks()` surface, so a manifest naming both must name an SDK whose types match. A caret would let a consumer resolve SDK 0.19.7 (on `trust-tasks-rs` 0.6) alongside `trust-tasks-rs` 0.9 — reintroducing precisely the split this PR removes.

Every other check passes, including `publish dry-run`, `release guards`, all `facade/*`, all `mediator/*`, all storage backends, and both Snyk scans.

PR #717, opened 2026-08-19
build(deps): move trust-tasks-rs to 0.11 across the messaging crates

stormer78:
## Why

`trust-tasks-rs` 0.11 carries the `vta/services/*` task family — the eight
specs that supersede the twenty `/services/*` REST routes on a VTA.

**The VTA can't use them while this workspace pins 0.9.** `vta-sdk` has to speak
the same version these crates do: `acl_setup.rs` builds a `MediatorAcl` from its
own `trust-tasks-rs` and hands it to `atm.trust_tasks().account_update(...)`, so
two semver-incompatible copies make that a type error rather than a link:

```
error[E0308]: expected `MediatorAcl`, found a different `MediatorAcl`
```

This is the upstream half of that move. Six crates go 0.9 → 0.11:
messaging-sdk, mediator, test-mediator, didcomm-service, helpers, text-client.

## The remaining duplicate is expected, and clears itself

A second `trust-tasks-rs` 0.9 stays in the lockfile, reached through the
mediator's `vta-sdk 0.25`. The workspace builds and tests clean with both
present — no `trust-tasks-rs` type crosses that boundary. It goes when VTI ships
a `vta-sdk` built on 0.11, which is exactly what this unblocks.

## This is the move the manifest already asked for

The note on `vta-sdk` in the mediator's `Cargo.toml` records the last time this
bit: a bounded `>=0.21.0, <0.23.0` range held the crate on 0.21.6 and kept a
stale `trust-tasks-rs` in the lockfile long after everything else had moved to
0.9. Its conclusion —

> A bounded range is a tax that comes due silently; prefer a caret and move it
> deliberately.

— is the same lesson, applied forward rather than in hindsight.

## Verification

```
cargo build --workspace                   clean
cargo clippy --workspace --all-targets    clean
cargo test --workspace                    3015 passing, 0 failures
```

One flake worth flagging, unrelated to this change:
`config_writer::tests::test_all_secret_storage_refs` failed on my first
workspace run and passed on the second, and passes in isolation both with and
without this change. It writes config, so parallel interference is the likely
cause — noting it rather than folding a fix into a dependency PR.

## After this

Publish `affinidi-messaging-sdk`, then VTI moves `vta-sdk` to
`trust-tasks-rs 0.11` and implements the `vta/services/*` handlers against the
**generated** types — with full schema-conformance coverage, which is the whole
reason for going this way round instead of hand-writing the payload bodies.
That matters here: the conformance sweep exists because two payload defects
shipped from hand-written types (#895 sending `null` for every unset member,
#919 sending `"mnemonic": null` on every call), both found in production.


stormer78:
## Both release guards now pass

Two follow-up commits, and each guard caught something real.

**`release guards` — four crates changed without a version bump.** A source
change that doesn't move the version never publishes, so the dependency bump
would have sat on main reaching nobody. Bumped as **patch** releases, following
this workspace's own precedent: `0.19.8` was itself a `trust-tasks-rs` move
(0.6.1 → 0.9.0, spanning three breaking releases) and shipped as a patch,
because none of the breakage reached the crate's source. Same here, and more so
— 0.10.0 and 0.11.0 are purely additive.

I did check whether it should be a minor: `account_update` takes `MediatorAcl`
in its public signature, so the types are exposed. The established reading here
is that what matters is whether the crate's own source has to change, and it
doesn't.

**`publish dry-run` — `didcomm-service` compiled against the published SDK.**
The dry-run resolves from crates.io by design, so it built against
`affinidi-messaging-sdk 0.19.8` (on trust-tasks-rs 0.9) while the crate itself
now uses 0.11 — the same `E0308` that motivated this PR, this time inside the
packaged tarball.

`check-publish-dry-run.sh` already anticipates exactly this shape and reports it
as *"waiting on a sibling this release publishes"* rather than a failure. It
recognises it **narrowly**, and correctly: the unmet requirement must name a
publishable crate in this workspace whose *local* version is exactly the version
required. Mine asked for `0.19.8` while the local SDK had moved to `0.19.9`, so
it read as ordinary registry drift — which is precisely what that guard exists
to catch. Aligning the five internal requirements to `0.19.9` makes the intent
legible to the guard and correct at release time.

## Verified locally on the committed tree

```
scripts/check-version-bumps.sh        All changed publishable crates were version-bumped.

scripts/check-publish-dry-run.sh      exit 0
  wait affinidi-messaging-didcomm-service — unresolvable only until a sibling in this release publishes first
  wait affinidi-messaging-mediator       — "
  wait affinidi-messaging-test-mediator  — "
  ok   affinidi-messaging-sdk
```

`cargo build --workspace` clean, `cargo test --workspace` 3015 passing.

PR #723, opened 2026-08-23
feat: add the did:webs DID method, with KERI-verified key state

stormer78:
Adds `did:webs` — `did:web` discovery with KERI-verified key state — as a new
method crate, and wires it into the resolver SDK and the `affinidi-tdk` facade
behind a `did-webs` feature.

Built on [`affinidi-keri-rs`](https://github.com/affinidi/affinidi-keri-rs)
0.3.0, which was published for this.

## Why it matters

A `did:webs` identifier ends in a KERI **AID** and publishes two artifacts:

| URL | contents |
|---|---|
| `https://<host>/<path>/<AID>/did.json` | the DID document, as a cache |
| `https://<host>/<path>/<AID>/keri.cesr` | the key event log it must be derived from |

The document this returns is **derived from the verified key event log**, never
copied from the published `did.json`. `did.json` carries no authority on its
own: it is cross-checked against the derivation, and a disagreement fails the
resolution rather than being resolved in either direction.

## `affinidi-did-webs` 0.1.0

| module | what it does |
|---|---|
| `identifier` | parsing and artifact URL derivation |
| `kel` | verifies the key event logs in a `keri.cesr` stream, including delegation |
| `document` | verified key state → DID document |
| `resolver` | the `did.json` cross-check |
| `fetch` | HTTPS retrieval with a size cap per artifact |

### Verified

- Every event's SAID, before any of its fields are trusted.
- Prior-event digest chaining and sequence ordering.
- Controller signatures, against the key set each event type is signed by.
- Pre-rotation commitments on every rotation.
- **Delegation** — a delegated event is accepted only if the delegator's own
  verified KEL, present in the same stream, anchors a seal naming that exact
  event. This is the `DelegatorAnchors` implementation that
  affinidi-keri-rs#5 deliberately left to the resolver, since a `Kever` cannot
  fetch or verify another identifier's KEL. Chains are bounded at 8 and
  cycle-checked: a crafted stream would otherwise spin a resolver in a loop.
- The published `did.json`: it must name this identifier (or its `did:web`
  twin) and publish **exactly** the keys the KEL authorised — no more, no fewer.

### Deliberately not verified, and documented as such in the README

- **TEL and ACDC state.** The designated-aliases attestation parses and its
  signing group is exposed, but nothing checks the registry behind it, so
  `alsoKnownAs` beyond the `did:web` twin is **not asserted**. A document from
  this crate therefore carries *fewer* `alsoKnownAs` entries than the published
  `did.json` may. That is a visible difference from other implementations and a
  deliberate choice: asserting an alias we cannot check is worse than omitting
  it.
- **Witness receipts** are parsed but not required; a KEL is accepted on
  controller signatures alone.
- **Service endpoints**, which come from KERI endpoint role authorisations.

### Some details worth knowing

- The AID is always the final path element, so unlike `did:web` a pathless
  `did:webs` never resolves to `/.well-known/`.
- `%3A` is decoded in either case, and compared normalised: RFC 3986 §6.2.2.1
  makes `%3a` and `%3A` equivalent, published artifacts write the port
  separator lowercase, and the specification writes it uppercase. Comparing raw
  strings rejects every real document.
- AID labels are validated against the CESR base64url alphabet. That is a
  syntax check, not a security one — the AID is proven by the KEL — but it is
  what keeps `../../etc/passwd` out of a fetch URL.
- EC point decompression goes through `affinidi-crypto`, which validates the
  point is on its curve. A compressed point is 33 bytes of server-supplied
  data, and an off-curve key must not reach a document.
- Artifacts are size-capped, checked against both the advertised
  `Content-Length` and the bytes actually read. A KEL grows with every rotation
  and has no natural bound.

## Wiring

`did:webs` is not modelled by `MethodName`, so it registers under
`MethodName::Other("webs")` and arrives at the resolver as
`DIDMethod::Other { method: "webs", .. }`. A typed variant would be a
source-breaking change to `MethodName`, which is not `#[non_exhaustive]`, and
buys nothing — dispatch is by name either way. A test pins the dispatch key,
because a drift between registration and lookup would degrade silently to
"unsupported method" rather than fail loudly.

## Two problems found along the way

**An unmodelled DID method was cached forever.** `DIDExpiry` decides expiry via
`DIDMethod::try_from(method)`, which returns `Err` for a method it does not
model rather than `OTHER` — so the documented "OTHER is mutable" branch was
unreachable from that path and such documents were given **no TTL at all**.
They now take the mutable TTL, which is the conservative direction: a document
fetched from infrastructure that can change must be re-fetched, or a key
rotation is never picked up. This affected **every custom-registered resolver**,
not just `did:webs`. Two tests pin it.

**`affinidi-cesr` was being built twice.** `affinidi-did-webs` depends on it by
version, and so does `affinidi-keri-core` from the registry, so the workspace
resolved both a local copy and a crates.io copy. A `Matter` crossing between
them is a mismatched type whose error points anywhere but at the cause.
`scripts/check-workspace-duplicates.sh` caught it; fixed with a
`[patch.crates-io]` redirect.

## Testing

Conformance tests run against the artifacts published for
`did:webs:did-webs-service%3a7676:ENro7uf0…`, reproduced from the Apache-2.0
`hyperledger-labs/did-webs-resolver` reference implementation with attribution.
These are the only bytes in the suite this workspace did not produce itself, so
they are the only ones that can catch us agreeing with ourselves rather than
with the ecosystem.

The negative cases are the substance: a tampered key event log, a `did.json`
naming another identifier, one publishing a key the KEL never authorised, one
omitting a key it did, and one lifted to a different host — a published
document names the identifier it belongs to, so it cannot be replayed elsewhere.

`cargo test -p affinidi-did-webs` — 29 passed.
`cargo test -p affinidi-did-resolver-cache-sdk --features did-webs` — 55 passed.
Per-package builds, `--features did-webs` on both the SDK and the facade,
clippy, `cargo fmt --check`, `check-workspace-duplicates.sh`,
`check-changelogs.sh` and `check-version-bumps.sh` all clean.

## Versions

`affinidi-did-webs` 0.1.0 (new), `affinidi-did-resolver-cache-sdk`
0.8.25 → 0.8.26, `affinidi-tdk` 0.8.5 → 0.8.6. Both bumps are additive and stay
within their current minor, since external consumers pin `"0.8"`.

> `affinidi-did-webs` is a new crate name, so its first publish has to be done
> by a human — CI's Trusted Publishing returns 403 for a crate that does not
> exist yet, and the failure cascades to the crates that depend on it.


stormer78:
## The two red checks are expected, and cannot pass before merge

`publish dry-run` and `verify-publish` both fail with exactly one error:

```
no matching package named `affinidi-did-webs` found
location searched: crates.io index
```

`affinidi-did-webs` itself packages fine — its own dry-run passed. What fails is
its two dependents, `affinidi-did-resolver-cache-sdk` and `affinidi-tdk`. Both
checks resolve each crate against the registry *in isolation*, and a crate added
in the same PR is not there yet. Nothing can make them green while the crate is
unpublished.

Every other check passes: the full test suite, clippy, cargo-deny, audit, fmt,
deprecation check, all seven facade feature combinations, the mediator and setup
matrices, release guards, and the Snyk scans.

## What has to happen at release time

This is **not** the usual "the release job publishes in dependency order, so it
resolves itself" case. `affinidi-did-webs` is a **new crate name**, and CI cannot
create one: crates.io Trusted Publishing returns 403 for a crate that does not
exist yet, and that failure cascades to every crate depending on it.

So before the first release that includes this:

```
cargo publish -p affinidi-did-webs
```

must be run **by a human**, once. Every release after that is CI's job as usual.
Without it, the release run fails on `affinidi-did-webs` and then on
`affinidi-did-resolver-cache-sdk` and `affinidi-tdk` behind it.

Same trap the `affinidi-keri-*` crates hit — worth checking this comment before
kicking off the release rather than after.

PR #737, opened 2026-08-23
feat(did-webs): derive service endpoints from signed KERI authorisations

stormer78:
The last of the four gaps `did:webs` shipped with. A document now carries
service endpoints — but only ones the AID actually signed for.

## Two signed replies, or nothing

```
/end/role/add   signed by THIS AID   "I authorise <eid> in role <role>"
/loc/scheme     signed by <eid>      "I am reachable at <url> over <scheme>"
```

Both halves are required:

- the first without the second is an **authorisation pointing nowhere**
- the second without the first is a **stranger volunteering** to act on
  someone's behalf

Neither is published. A `did:webs` document does not get to list whatever
endpoints it likes, any more than it gets to assert its own aliases.

`/end/role/cut` withdraws an authorisation. Where the same subject is addressed
more than once the latest `dt` wins — so a stale `cut` replayed after an `add`
does **not** take effect, which is pinned in both directions.

## Signatures, both ways identifiers come

- **Transferable** — indexed signatures in a transferable signature group,
  checked against the key state that identifier's own KEL establishes. That KEL
  must be in the stream and must verify.
- **Non-transferable** (a witness, typically) — the public key *is* the prefix,
  so a receipt couple is checked directly.

Roles are deliberately **not** enumerated. KERI does not fix the set, and
rejecting an unknown role would silently drop endpoints we have simply not heard
of.

## Shape

Matches the reference implementation
(`hyperledger-labs/did-webs-resolver`, `didding.py`):

```json
{"id": "#<eid>/<role>", "type": "<role>", "serviceEndpoint": {"http": "http://…"}}
```

## Tests

The published artifact carries no `rpy` messages, so these build them with real
keys and real signatures. The negative cases are the substance:

| test | asserts |
|---|---|
| `an_authorised_endpoint_becomes_a_service` | the happy path, end to end into the document |
| `a_later_cut_withdraws_the_authorisation` | `cut` supersedes |
| `an_earlier_cut_does_not_withdraw_a_later_add` | a stale `cut` replayed afterwards does not |
| `an_authorisation_for_another_aid_is_ignored` | `cid` must be this AID |
| `an_unsigned_authorisation_is_ignored` | no signature, no service |
| `an_endpoint_that_never_said_where_it_is_publishes_nothing` | a service with no endpoint is worse than none |
| `a_location_signed_by_someone_else_is_ignored` | a third party naming where another identifier lives is a redirection, not a location |

## Versions

`affinidi-did-webs` 0.3.0 → **0.4.0** (`document_from_keys` takes the derived
services), `affinidi-did-resolver-cache-sdk` 0.8.28 → 0.8.29, `affinidi-tdk`
0.8.8 → 0.8.9.

## Verification

`cargo test -p affinidi-did-webs` — **47 passed** (19 unit, 16 conformance, 8
service, 4 witness), up from 39. clippy, `cargo fmt --check`, and all three repo
guards clean.

Note this branches from `main` at #735, so it does not include #736
(`did:scid:ke`). The two touch different crates and can merge in either order —
whichever lands second will want its `affinidi-did-webs` requirement checked, as
#736 pins `0.3`.

### DIFF
--- a/crates/identity/affinidi-did-resolver-cache-sdk/CHANGELOG.md
+++ b/crates/identity/affinidi-did-resolver-cache-sdk/CHANGELOG.md
@@ -1,8 +1,10 @@
 # Affinidi DID Resolver Cache SDK
-## 0.8.29
+## 0.8.30
 
 ### Changed
 
+- `did-webs` feature now requires `affinidi-did-webs` 0.4, which derives service
+  endpoints from signed KERI endpoint authorisations.
 - `did-scid` feature now requires `did-scid` 0.2, which adds `did:scid:ke:1`
   (KERI AIDs via `did:webs`) and reports an unresolvable format by name.
 
@@ -122,7 +124,9 @@ called out here rather than encoded in the version — see R3.6 in CLAUDE.md.
 
 ## 23rd July 2026
 
-### 0.8.21 — `resolve_any` accepts a community name
+#
+
+## 0.8.21 — `resolve_any` accepts a community name
 
 `resolve_any("example.com/@")` now classifies as an agent name and resolves,
 where it previously failed as malformed. The agent name FAQ gives a name with
--- a/crates/identity/affinidi-did-resolver-cache-sdk/Cargo.toml
+++ b/crates/identity/affinidi-did-resolver-cache-sdk/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "affinidi-did-resolver-cache-sdk"
-version = "0.8.29"
+version = "0.8.30"
 description = "Affinidi DID Resolver SDK"
 edition.workspace = true
 authors.workspace = true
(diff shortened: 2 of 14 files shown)


stormer78:
Rebased onto `main` after #736. Conflicts resolved.

Both PRs bumped the same two crates to the same versions, so the collision was
in `affinidi-tdk` and `affinidi-did-resolver-cache-sdk` — their manifests and
changelogs. Resolved by keeping **both** dependency bumps rather than either
side's, and renumbering this PR's releases past the ones #736 already published:

| crate | was | now |
|---|---|---|
| `affinidi-did-resolver-cache-sdk` | 0.8.29 | **0.8.30** |
| `affinidi-tdk` | 0.8.9 | **0.8.10** |

One cross-crate consequence git could not have flagged: **#736 shipped
`did-scid` 0.2.0 pinning `affinidi-did-webs = "0.3"`**, and this PR moves that
crate to 0.4.0. Left alone the workspace would need both versions at once and
fail to resolve. `did-scid` now requires `"0.4"` and goes to **0.2.1** with a
changelog entry.

Re-verified after the rebase: `affinidi-did-webs` 47 tests and `did-scid` 29
tests pass, `did-scid --features did-webs` builds, clippy and
`cargo fmt --check` clean, and all three repo guards pass with all four version
bumps documented.

PR #751, opened 2026-09-01
feat!: move affinidi-mdoc to coset 0.4

stormer78:
Unblocks
[OpenVTC/verifiable-trust-infrastructure#1225](https://github.com/OpenVTC/verifiable-trust-infrastructure/pull/1225),
which cannot move on its own.

## Why the consumer is stuck

`coset` is in `affinidi-mdoc`'s **public API**, not an implementation detail:

- `IssuerSigned::issuer_auth` and `Document::issuer_auth` are `pub` fields of
  type `coset::CoseSign1`
- `sign_mso` returns one, `extract_kid` takes one
- `CoseKey::default_algorithm` returns `Option<coset::iana::Algorithm>`

So which `coset` this crate compiles against is part of what it promises. VTI's
`vta-vault` names those types directly — it compares an `alg` taken from
`IssuerSigned` against one it builds from its own `coset` — and moving only its
side produced:

```
error[E0308]: mismatched types
   --> vta-vault/src/receive.rs:439:29
    |
439 |     if alg.as_ref() != Some(&expected) {
    |                             ^^^^^^^^^ expected `RegisteredLabelWithPrivate<Algorithm>`
    |
note: there are multiple different versions of crate `coset` in the dependency graph
```

with `affinidi-mdoc 0.2.7` still holding `coset 0.3.8` underneath it. The two
sides have to move together; this is the half that has to move first.

## What's in the change

**No source change anywhere.** The API this workspace uses is identical across
the bump. coset 0.4.0's one breaking change is the `crit` field on `Header`
gaining support for private-use labels ([RFC 9052
§3.1](https://datatracker.ietf.org/doc/html/rfc9052#name-common-cose-header-paramete)),
which nothing here touches; 0.4.1 and 0.4.2 are additive. `affinidi-mdoc`
compiled unmodified on the first try and all 126 of its tests pass.

Versions, and why each:

| crate | bump | why |
|---|---|---|
| `affinidi-mdoc` | 0.2.7 → **0.3.0** | `coset` is in its public API, so the bump is breaking |
| `affinidi-tdk` | 0.10.0 → **0.11.0** | `pub use affinidi_mdoc as mdoc` — the break is in its public API too |
| `affinidi-tdk-test-support` | stays **0.8.3** (unreleased) | `coset` is internal here; entry joins the dependency refresh already in that section |

Plus the four `affinidi-tdk = "0.10"` requirements in `crates/messaging/*`. A
published `affinidi-tdk` 0.10.0 stays pinned to `affinidi-mdoc` 0.2 and can
never reach 0.3, so leaving those would strand the consumer this exists to
unblock.

**Minor rather than patch** on the two published crates, deliberately. A patch
bump would let cargo resolve a consumer on `coset 0.3` beside `affinidi-mdoc` on
0.4 and fail with duplicate-type errors instead of a version conflict — the
exact failure being fixed here, reintroduced one level down.

[ADR 0003](../blob/main/docs/adr/0003-public-api-semver-policy.md) point 3's
patch-bump carve-out does not apply. That exists for crates redirected through
`[patch.crates-io]`, where a minor bump breaks the redirect; `affinidi-mdoc` is
in no patch section here, and the consuming workspace removed its own.

## Testing

`cargo check --workspace --all-targets`, `cargo clippy --workspace
--all-targets` and `cargo fmt --all` clean. `affinidi-mdoc` (126),
`affinidi-tdk-test-support` and `affinidi-tdk` tests all pass. `Cargo.lock` now
resolves exactly one `coset`, at 0.4.2.

## After this merges

VTI #1225 is a straight re-run once `affinidi-mdoc 0.3` is published — the bump
there needs no code change either, it just needs both sides on one `coset`.

### DIFF
--- a/crates/credentials/affinidi-mdoc/CHANGELOG.md
+++ b/crates/credentials/affinidi-mdoc/CHANGELOG.md
@@ -1,5 +1,36 @@
 # Affinidi mdoc Changelog
 
+## 1st September 2026 (0.3.0)
+
+Moves to **coset 0.4**.
+
+No source change — the API this crate uses is identical across the bump, and
+0.4.0's one breaking change (the `crit` field on `Header` gains support for
+private-use labels, per [RFC 9052
+§3.1](https://datatracker.ietf.org/doc/html/rfc9052#name-common-cose-header-paramete))
+touches nothing here. All 126 tests pass unmodified.
+
+**Minor rather than patch, because `coset` is in this crate's public API.**
+`IssuerSigned::issuer_auth` and `Document::issuer_auth` are `pub` fields of type
+`coset::CoseSign1`; `sign_mso` returns one; `extract_kid` takes one; and
+`CoseKey::default_algorithm` returns `Option<coset::iana::Algorithm>`. A
+consumer that names those types has to move in the same step, so the version
+has to say so — a patch bump would let cargo resolve a consumer on `coset 0.3`
+beside this crate on 0.4 and fail with duplicate-type errors rather than a
+version conflict.
+
+That is not hypothetical. It is what blocked
+[OpenVTC/verifiable-trust-infrastructure#1225](https://github.com/OpenVTC/verifiable-trust-infrastructure/pull/1225):
+`vta-vault` compares an `alg` taken from `IssuerSigned` against one it builds
+from its own `coset`, and moving only its side produced
+
+```
+error[E0308]: mismatched types
+note: there are multiple different versions of crate `coset` in the dependency graph
+```
+
+This release is what lets that side move.
+
 ## 16th August 2026 (0.2.7)
 
 Adds the CBOR wire codec for **`DeviceResponse`** — `to_cbor_bytes` /
--- a/crates/credentials/affinidi-mdoc/Cargo.toml
+++ b/crates/credentials/affinidi-mdoc/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "affinidi-mdoc"
 description = "ISO/IEC 18013-5 mdoc (Mobile Document) implementation for mobile driving licences and eIDAS attestations."
-version = "0.2.7"
+version = "0.3.0"
 edition.workspace = true
 authors.workspace = true
 readme = "README.md"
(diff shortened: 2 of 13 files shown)


stormer78:
CI update — 42 pass, 1 fail, and the remaining one is the known-red for this class of change.

**Fixed since the first push (two real problems):**

- `release guards` was right and I was wrong: `affinidi-messaging-didcomm-service`, `affinidi-messaging-test-mediator` and `affinidi-tdk-test-support` all had manifests edited without a version bump. Now 0.5.1 / 0.4.3 / 0.8.4, with each crate's existing `Unreleased` heading rolled forward — those older numbers were never published, so everything ships under the new one. (I had misread test-support's `Unreleased (0.8.3)` as meaning it was already bumped; the guard compares against the PR base, and main is at 0.8.3 too.)
- `publish dry-run` was failing for a **bug in the guard itself**, now fixed in the second commit. `awaits_sibling_release` compared the requirement cargo reports against the crate's local version with string equality, and internal requirements here are written `major.minor` by convention — so `req_ver` was `0.11` while `local_ver` was `0.11.0`, `[ "0.11.0" = "0.11" ]` was false, and every atomic multi-crate bump reported FAIL. That made the documented `wait` path ("ONE FAILURE IS NOT A FAILURE") unreachable for the only shape it exists to recognise. Now prefix-with-boundary; the `.` matters, or `0.3` would also match a local `0.30.0`.

That guard now reports what it should, exit 0:

```
  ok   affinidi-mdoc
  wait affinidi-messaging-didcomm-service — unresolvable only until a sibling in this release publishes first
  wait affinidi-messaging-test-mediator
  wait affinidi-tdk-test-support
  wait affinidi-tdk
All changed publishable crates pass cargo publish --dry-run.
```

**The one still red: `rust-pipeline / verify-publish / Verify Publishable`.**

Same structural cause — `affinidi-mdoc "^0.3"` is not on crates.io yet — but this is the older guard, and it has no `wait` concept: it marks FAILED and collects the crate. It cannot be satisfied before merge, because the dependency it wants only exists once this merges and publishes.

This is the accepted state for a coordinated breaking release here: **#743 and #744 both merged with these same two checks red.** This PR is strictly better than those — `publish dry-run` is green here where it was red on both.

Worth noting the two guards now overlap: `publish-dry-run.yaml` does everything `verify-publish` does and more (default features, `-D warnings`, changed-crate scoping) *and* understands the ordered-publish case. If you want, I'll teach `verify-publish` the same `version_satisfies` check, or retire it in favour of the newer one — happy either way, but I've left it out of this PR since it's beyond the coset bump.

PR #753, opened 2026-09-03
Harden fuzz CI against supply-chain risk (SEC-3898/99)

roland-gh-aff:
Brings the tdk-rs fuzz workflow to the same supply-chain posture as webvh, per the dogwood advisory review.

## fuzz.yaml (SEC-3898, SEC-3899)
- Commit Cargo.lock for both fuzz subcrates (affinidi-messaging-didcomm, affinidi-sd-jwt) and enforce with `cargo fetch --locked` plus an offline build.
- Restrict runner egress with harden-runner (crates.io, rust, github, apt only).
- Pin cargo-fuzz to 0.13.2 and SHA-pin all workflow actions.
- permissions and the gnu build target were already correct.

Verification: green in CI, 6 of 6 targets.

Also includes the earlier dogwood commit on this branch: block SSRF-prone hosts and cap response body in the did-web resolver.

Tickets: SEC-3898, SEC-3899.


stormer78:
Reviewed this and pushed fixes for what came out of it (2 commits). Summary of what changed and why.

## did:web SSRF guard — four holes and an over-claim

The guard as written classified the host **as spelled in the DID**, and only some spellings.

1. **`did:web:localhost.` walked straight through.** A trailing root dot is a legal hostname and survives URL normalisation, so `host_str()` returns `"localhost."`, `h == "localhost"` never matched, and the request was issued (verified: connection-refused, i.e. it resolved to loopback and connected). Same for `*.localhost.` / `*.local.`. Now normalised before the name comparisons.

2. **CGNAT `100.64.0.0/10` was not blocked**, so `did:web:100.100.100.200` — the Alibaba/Oracle metadata endpoint — was reachable. That range is also the node/pod CIDR in many EKS/GKE deployments. Added, along with `0.0.0.0/8`, `192.0.0.0/24` and `198.18.0.0/15`.

3. **IPv4-compatible IPv6 slipped past.** Only `to_ipv4_mapped()` was unwrapped, so `::ffff:127.0.0.1` was caught but `::127.0.0.1` was not. `to_ipv4()` covers both. NAT64 `64:ff9b::/96` now unwraps its embedded v4, and `fec0::/10` is blocked.

4. **Any DNS name defeated the guard.** No rebinding needed: register `evil.example.com` with a static A record of `169.254.169.254`, publish `did:web:evil.example.com`, and the resolver fetches the metadata endpoint. The doc comment said rebinding was "a connect-time concern handled elsewhere" — there was no elsewhere, in this crate or in `network_resolvers.rs`, and the CHANGELOG's "stops a did:web value from steering the resolver at internal endpoints" over-claimed.

   Rather than downgrade the claim, the default client now installs a DNS resolver that refuses any name resolving to a non-routable address. Because the connection is made to the addresses that resolver returns, this closes the rebinding window too — there is no second, unchecked lookup between guard and connect. The refusal travels out through reqwest's error chain and is recovered as `BlockedHost` rather than a generic transport failure; there is a test asserting that specifically, since it depends on reqwest's wrapping.

Also: the URL parse was `if let Ok(...) && ...`, so a parse failure **skipped the check**. Not exploitable today (the client re-parses with the same parser and errors), but a security control shaped as "skip on error" is one refactor from being a real bypass. It is now fatal, and the parsed `Url` is handed to the client instead of re-parsing the string.

## The behaviour break needed an escape hatch

0.1.4 turned every did:web host on RFC-1918, `.local` or loopback into an unconditional `BlockedHost` — an internal mediator resolving peers on `10.x`, or this repo's own `did:web:localhost%3A8080` test identity — with no opt-out and no CHANGELOG call-out, shipped as a patch. The only escape was pinning `=0.1.3`.

Added `HostPolicy::{PublicOnly, AllowPrivate}` with `DIDWeb::with_policy` / `with_client_and_policy`, and `WebResolver::with_policy` in the cache-sdk (0.8.34 → 0.8.35) so the hatch is actually reachable from the resolver stack — every consumer in this repo goes through `WebResolver::new()`. `guarded_dns_resolver()` is public so a caller passing their own client to `with_client` can install the same protection; without it they get only the literal-address half, which is now documented. Both CHANGELOGs lead with the behaviour break, and `docs/testing/docker-compose.md` says the test identity no longer resolves by default and how to opt in.

16 tests (was 9), clippy clean, workspace builds.

## Fuzz CI

**`cargo fetch --locked` turns every version bump of a fuzzed crate into a red fuzz job.** The fuzz crates depend by path on the crates they fuzz, and a path package still carries a version in `Cargo.lock`. `check-version-bumps.sh` *requires* a bump on any source edit to exactly those crates, and this workflow triggers on exactly those paths. Reproduced: bump `affinidi-sd-jwt`, and the gate fails with `cannot update the lock file ... because --locked was passed` — for a reason that has nothing to do with supply-chain risk. Worse for `affinidi-crypto`, which the targets pull in transitively and which does not match the path filter: the breakage lands unseen and surfaces in the 03:17 nightly.

Replaced with `scripts/check-fuzz-lock.sh`, which checks what the gate actually means: re-resolve, then require every **crates.io** version and checksum to be exactly as committed. Path packages are free to renumber; a new or changed registry version still fails, with a diff, and must be reviewed and committed. Verified three ways — passes clean, passes after a version bump where the old gate fails, fails on a moved registry pin. It restores the committed lock, so no working-tree side effects.

**harden-runner was blocking the crash-artifact upload.** The allowlist covered the build path but not `results-receiver.actions.githubusercontent.com:443` or `*.blob.core.windows.net:443`, which `actions/upload-artifact@v4` needs. The upload step is `if: failure()`, so a green run never exercises it — the first time a target actually crashed, the reproducer would have been lost to an egress block rather than uploaded. Both endpoints added.

## Not changed, worth a decision

The PR path filter still excludes `crates/core/affinidi-crypto/**`. A crypto change no longer breaks the gate, but it also isn't fuzz-smoke-tested per-PR. Adding it means six fuzz jobs on every crypto PR — deliberate call rather than something to fold into a review fix.

## Checked and clear

- `!Cargo.lock` in the two fuzz `.gitignore`s does correctly override the root pattern (no parent directory is excluded).
- The `fuzz/*|*/fuzz/*` exclusion in `check-version-bumps.sh` is right — the fuzz crates are standalone workspaces, invisible to `cargo metadata --no-deps`, so their files previously bubbled up to the publishable parent. No crate has a `src/fuzz/` the pattern would over-match.
- SHA-pinning `dtolnay/rust-toolchain # nightly` is fine despite that action deriving its toolchain from `github.action_ref`, because the job sets `RUSTUP_TOOLCHAIN: nightly`.
- The 1 MiB `chunk()` body cap is correct — inclusive boundary, no unbounded buffering.
- The numeric-obfuscation vectors (`2130706433`, `0177.0.0.1`, `0x7f.1`, `127.0.0.1.`) were already caught: the `url` crate normalises them to `127.0.0.1` before the guard sees them.
- The red `checks` run on this branch is unrelated: `cargo-audit v0.22.0` fails to build against `tinyvec 1.13.0` (`cannot find macro vec`), plus a pre-existing missing `scripts/annotate-audit-dependency-type.sh`.

PR #756, opened 2026-09-05
fix(mediator): bind a TSP envelope's sender to the authenticated session

### LINKED ISSUE #754: Messaging mediator: envelope sender VID differing from session DID is forwarded. Intended contract?
Measured on the public dev mediator (Dogwood, webvh.storm.ws): a TSP frame sent on a socket authenticated as DID A, with envelope sender DID E (never authenticated), is forwarded normally. Repro: https://github.com/real-life-org/rltp-spec/blob/main/binding/.sender-hypothesis.mjs

Not a bug report: this is consistent with TSP's intermediary posture, and our transport binding relies on it (separate connection vs. egress identities). Question: stable property or future anti-spoofing target? If the latter, we'd rather switch to routed mode now than break silently later.

stormer78:
Closes #754.

## What was reported

Measured against a public deployment: a TSP frame sent on a socket authenticated as DID **A**, carrying envelope sender DID **E** which never authenticated, is forwarded normally. The reporter asked whether that is a stable property or a future anti-spoofing target.

It is neither intended nor stable — it is a gap.

## Why it is a gap, not a posture

A TSP envelope names its sender in the clear and the mediator does not decrypt it, so `meta.sender` is a *claim*. `handle_inbound_tsp` never consulted the session at all: it passed that claim straight to `deliver_opaque`, which hashes it into the recipient's `delivery_decision` access-list lookup.

Three things say this was missed rather than decided:

1. **`docs/acls.md` §6 already promised the check.** The flow is headed *"Direct delivery (DIDComm and TSP)"* and lists `force_session_did_match` as a step. The contract was stated; only the implementation was missing.
2. **It is the same bug we already fixed once.** Mediator 0.15.5 (May 2026) fixed exactly this on the DIDComm path — *"an authenticated session could set `skid` to any allow-listed DID and bypass both checks"*. The TSP path was written after that fix and did not carry it.
3. **The function's own comment asserts the invariant it did not hold** — `"The mediator routes on the authenticated sender"`, where the value is unauthenticated.

No test pinned the old behaviour, so closing it breaks no pinned expectation.

## What it cost

The claimed sender feeds the recipient's access list, so a client could borrow an allow-listed VID and be admitted to an inbox that does not admit it. The recipient still fails to unpack end-to-end, so this is an access-list / flooding bypass rather than identity forgery — which is precisely what the 0.15.5 fix was about.

`spoofed_sender_cannot_pass_the_recipients_access_list` demonstrates it rather than asserting it: revert just the guard and that test fails **at the spoofed send succeeding**.

## The fix

- Bind the cleartext envelope sender to the session DID under the same `security.force_session_did_match` switch, reusing the existing `check_direct_delivery_session_match` predicate so both protocols return the same `authorization.did.session_mismatch` report.
- Checked on the **outer** envelope, before the receiver branch, so it covers relayed and nested submissions: a client must have authored the layer it hands over. Inner layers are exempt by construction — sealed to someone else.
- **Anonymous sessions exempt**, exactly as on the DIDComm side since 0.20.4 (#748). An inter-mediator relay hop is POSTed to `/inbound` with no `Authorization` header and lands on `ANON-INBOUND`, whose DID is empty. Without the exemption, cross-mediator TSP delivery stops entirely. Residual cost is the one already named for blind relay: on an anonymous hop the claimed sender stays unverified — inherent to relaying.
- Direct TSP delivery also now checks the sender's own `SEND_MESSAGES`. That is load-bearing on the **WebSocket** ingress in particular, which gates only on `LOCAL` at upgrade: a DID whose `SEND_MESSAGES` had been revoked could still post TSP frames over a socket.

## Deliberately not fixed here

TSP direct delivery still ignores `security.local_direct_delivery_allowed`, so a deployment that turned direct delivery off to force everything through a routing envelope still accepts direct TSP.

I implemented that gate, measured it, and backed it out: it fails **~20 existing tests across nine files**, all because the fixture defaults the flag off and the TSP tests were written against a path that ignored it. That is blast radius deserving its own change and changelog entry, not something to fold into a security fix where a reviewer cannot tell which test edits are incidental. `docs/acls.md` now names the deviation instead of implying it does not exist. Follow-up issue to come.

## Behaviour change (R3.6)

A client that deliberately sends under a VID other than the one it authenticated as is now refused with `e.p.authorization.did.session_mismatch`, unless the deployment sets `security.force_session_did_match = "false"`.

Deployments relying on the split between connection identity and egress identity — which is what #754 describes — should move to TSP **routed** mode, where the outer sender is the connection identity and the egress identity travels inside the sealed layer. That shape satisfies the binding by construction and is what survives this class of fix.

Mediator bumped to 0.20.7 with a changelog entry.

## Testing

- New `tsp_session_binding.rs` (4 tests): the regression, a matching-sender control, the access-list bypass demonstration, and one pinning that `force_session_did_match = false` still accepts a mismatched sender.
- Reverting only the guard fails exactly the two session-binding tests and nothing else.
- Full `affinidi-messaging-test-mediator --features tsp` suite green, cross-mediator forwarding included; `affinidi-messaging-mediator` unit tests green (239 + integration).
- `cargo fmt --check`, clippy `-D warnings` on `didcomm,tsp,memory-backend`, `check-version-bumps.sh`, `check-changelogs.sh`, `check-workspace-duplicates.sh` all clean.

## Pre-merge checklist (vti-stack-development-guide)

- **R1.1** — unaffected; this path returns a problem report on refusal, never a silent `Ok`.
- **R3.6** — behavioural change to send semantics, called out above and in CHANGELOG; consuming repos need to know before upgrading.
- **R6.2 / R1.7** — untouched.

### DIFF
--- a/crates/messaging/affinidi-messaging-mediator/CHANGELOG.md
+++ b/crates/messaging/affinidi-messaging-mediator/CHANGELOG.md
@@ -1,5 +1,62 @@
 # Changelog
 
+## Unreleased (0.20.7) — bind a TSP envelope's sender to the authenticated session
+
+**Security fix: the TSP ingress trusted the envelope's claimed sender.**
+Reported as [#754], measured against a public deployment: a TSP frame sent on a
+socket authenticated as DID A, carrying envelope sender DID E (never
+authenticated), was forwarded normally.
+
+A TSP envelope names its sender in the clear and the mediator does not decrypt
+it, so `meta.sender` is a claim. `handle_inbound_tsp` never consulted the
+session at all — it passed that claim straight to `deliver_opaque`, which hashes
+it into the recipient's `delivery_decision` access-list lookup. An authenticated
+client could therefore borrow any allow-listed VID and be admitted to an inbox
+that does not admit it.
+
+This is the TSP twin of the DIDComm direct-delivery bypass fixed in 0.15.5. The
+TSP path was written after that fix and did not carry it. `docs/acls.md` §6 has
+documented the flow as "Direct delivery (DIDComm and TSP)" — including the
+session-DID match — the whole time, so the contract was already stated; only the
+implementation was missing.
+
+- The cleartext envelope sender is now bound to the session DID under the same
+  `security.force_session_did_match` switch the DIDComm paths use, reusing the
+  same `check_direct_delivery_session_match` predicate and returning the same
+  `authorization.did.session_mismatch` problem report.
+- Checked on the **outer** envelope, before the receiver branch, so it covers
+  relayed and nested submissions too: a client must have authored the layer it
+  hands over. Inner layers are exempt by construction — they are sealed to
+  someone else.
+- **Anonymous sessions are exempt**, exactly as on the DIDComm side since 0.20.4:
+  an inter-mediator relay hop is POSTed to `/inbound` with no `Authorization`
+  header and lands on the anonymous `ANON-INBOUND` session, whose DID is empty.
+  Without the exemption cross-mediator TSP delivery would stop entirely. The
+  residual cost is the one already named for blind relay: on an anonymous hop the
+  claimed sender stays unverified, which is inherent to relaying.
+- Direct TSP delivery now also checks the sender's own `SEND_MESSAGES`, mirroring
+  the DIDComm direct-delivery branch. This is load-bearing on the **WebSocket**
+  ingress in particular, which gates only on `LOCAL` at upgrade: a DID whose
+  `SEND_MESSAGES` had been revoked could still post TSP frames over a socket.
+
+**Behaviour change.** A client that deliberately sends under a VID other than the
+one it authenticated as will now be refused with
+`e.p.authorization.did.session_mismatch` unless the deployment sets
+`security.force_session_did_match = "false"`. Deployments relying on the split
+between connection identity and egress identity should move to TSP **routed**
+mode, where the outer sender is the connection identity and the egress identity
+travels inside the sealed layer — that shape satisfies the binding by
+construction (R3.6: coordinate with consuming repos).
+
+**Still outstanding, tracked separately:** TSP direct delivery does not honour
+`security.local_direct_delivery_allowed`, so a deployment that has turned direct
+delivery off to force everything through a routing envelope still accepts direct
+TSP. That is a policy change with far wider blast radius than this fix — it
+changes the default fixture's behaviour for ~20 existing tests — and is being
+handled on its own rather than folded in here.
+
+[#754]: https://github.com/affinidi/affinidi-tdk-rs/issues/754
+
 ## Unreleased (0.20.6) — TSP forwarding follows a next hop that names its mediator by DID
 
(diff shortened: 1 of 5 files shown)


stormer78:
## Triage of the three AgenticSec findings

All three are true positives. None of them is a defect in this change — two are accepted trade-offs the PR already documents, one is the gap this PR deliberately split out. One of them did expose a real omission in how I worded the trade-off, now fixed in 971d7c3.

### Alert 66 — `force_session_did_match = false` still permits spoofing (medium, CWE-863)

**Accurate, working as intended.** The switch gates the DIDComm binding in exactly the same way (`inbound.rs:761`, `:835`), and putting TSP under the same switch was the point: operators reason about one flag, not one per protocol. Making the TSP binding unconditional would also strand the reporter of #754, who currently depends on the split and needs a migration window — removing their only lever while telling them to migrate is not a kindness.

The bot's own triage reaches the same place: *"the precondition is a non-default, explicit operator misconfiguration."* The default is `true`, and boot-time validation already refuses `force_session_did_match = true` alongside `block_anonymous_outer_envelope = false`.

Won't fix. The residual risk is the documented cost of an escape hatch that exists so this fix can ship without breaking people.

### Alert 67 — anonymous relay hops carry unverified sender claims (medium, CWE-290/345)

**Accurate, and it caught a real omission in my wording.**

The exemption itself stands: the inter-mediator forward POSTs to `/inbound` with no `Authorization` header (`mediator-common/src/tasks/forwarding/processor.rs`), so the hop lands on `ANON-INBOUND` with an empty DID. Without the exemption, cross-mediator TSP delivery stops entirely — that was the #748 regression.

What I got wrong was calling the residual cost *"the one already named for blind relay"* and leaving it there. That invites the inference that TSP inherits DIDComm's mitigation along with its risk. It does not: `RelayMode::Rewrap` + `processors.forwarding.relay_trusted_mediators` run through `relay_peer_trusted`, which is `#[cfg(feature = "didcomm")]` and referenced only from the DIDComm branches. **A TSP deployment that needs its relaying peer authenticated has no lever to pull today.**

Fixed in 971d7c3 (comment + changelog now state the asymmetry, and note that anonymous inbound is itself opt-in via `enable_inter_mediator_relay`, which bounds exposure to relay-enabled deployments). Filed as **#758** for the actual control.

### Alert 68 — TSP ignores `local_direct_delivery_allowed` (medium, CWE-862/284)

**Accurate, and already tracked as #757.** Pre-existing, not introduced here. I implemented the gate, measured it at ~20 failing tests across nine files — the fixture defaults the flag off and every TSP test was written against a path that ignored it — and backed it out rather than bury a security fix under a mechanical test migration where a reviewer could not tell which edits were incidental. `docs/acls.md` was amended in this PR to name the deviation instead of implying it does not exist.

Independent confirmation of the split is useful signal, and it raises the priority of #757.

---

Net: one wording fix pushed, two follow-ups tracked (#757, #758), no change to the fix itself. Worth noting the bot found nothing wrong with the binding, the anonymous exemption, or the `SEND_MESSAGES` addition — the three things most likely to have broken federation.

PR #784, opened 2026-09-09
chore(deps)!: trust-tasks-rs 0.18 -> 0.19 across the messaging family

stormer78:
Six crates in this family pin the Trust Tasks payload types, and they have to move as a set.
A mixed graph is what produces `expected MediatorAcl, found a different MediatorAcl` — the
failure this workspace has hit twice and now documents in two `Cargo.toml` comments:

> …this crate compiled against the *published* SDK rather than the workspace one — which is
> how a third `trust-tasks-rs` node survived the 0.12 move and produced
> `expected MediatorAcl, found a different MediatorAcl` from inside this workspace.

## Why now: the mixed graph has escaped this repo

`verifiable-trust-infrastructure` needs `trust-tasks-rs 0.19` to move to
`rooms/keys/present/0.2` and `rooms/owner/issue-authority/0.2`, and **cannot take it** while
`affinidi-messaging-sdk 0.22.0` requires `0.18`:

```
affinidi-messaging-sdk 0.22.0  ->  trust-tasks-rs 0.18.9
verifiable-trust-infrastructure ->  trust-tasks-rs 0.19.0
```

Pinning the consumer at 0.19 puts two `trust-tasks-rs` nodes in *its* graph, and `vta-sdk`
breaks on precisely the documented error, at the `vta-sdk` / `affinidi-messaging-sdk`
boundary. The compiler names the cause outright: *"there are multiple different versions of
crate `trust_tasks_rs` in the dependency graph"*.

Worth noting the shape: **nothing in the consumer's manifests is wrong**, and no semver check
would flag it. The break arrives through a public dependency, which is the class of change
`cargo-semver-checks` cannot see. The only fix is here.

## Is 0.19 safe for this family?

Yes — it is additive for everything these crates touch. `0.19` adds two room spec versions
and retires their predecessors, and a retired spec keeps its generated module, so no payload
type in use here changed shape. Verified by building and testing the workspace rather than by
reading the changelog.

## One duplicate node remains, deliberately

```
vta-sdk 0.32.4 (published) -> affinidi-tdk 0.11.0 -> affinidi-messaging-sdk 0.21.1 -> trust-tasks-rs 0.17.10
```

An older published copy of this workspace's own crate arriving back through a consumer. It is
untouched because it is a different problem and it compiles: no type crosses that boundary —
the mediator's `vta-sdk` dependency is feature-gated and hands nothing across. It is the cycle
the mediator's `Cargo.toml` comment already describes, and closing it means moving that
`vta-sdk` pin, which is a separate decision with its own history in that comment.

## Verified

- `cargo check --workspace --all-targets` — clean
- `cargo clippy --workspace --all-targets` — clean
- `cargo test --workspace --lib` — 1074 passed, 0 failed
- Lockfile: `trust-tasks-rs 0.18.9` is gone; `0.19.0` and the pre-existing `0.17.10` remain

Downstream: OpenVTC/verifiable-trust-infrastructure#1360 carries the dtg-credentials half and
notes this as its blocker; the `0.2` cut-over lands once this publishes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,38 @@ Per-crate version history is summarised here; for the full code history see
 
 ## [Unreleased]
 
+### Changed
+
+- **`trust-tasks-rs` 0.18 → 0.19 across the messaging family (breaking).**
+  **`affinidi-messaging-sdk` 0.23.0**, **`affinidi-messaging-didcomm-service` 0.7.0**,
+  **`affinidi-messaging-mediator` 0.23.0**, **`affinidi-messaging-test-mediator` 0.6.0**,
+  **`affinidi-tdk` 0.13.0**.
+
+  `trust-tasks-rs` is a **public** dependency of `affinidi-messaging-sdk` — types like
+  `account::update::v0_1::MediatorAcl` appear in its signatures — so moving it changes
+  these crates' APIs even though not a line of their own source did. That is why the bump
+  is a minor rather than a patch, and why `affinidi-tdk` moves with them: it re-exports the
+  SDK, and a consumer reaching `MediatorAcl` through the facade sees the same change.
+
+  These six move together because a **mixed** graph is the failure mode. Two `Cargo.toml`
+  comments in this workspace already record the symptom — `expected MediatorAcl, found a
+  different MediatorAcl` — from the two previous times a stale node survived a move.
+
+  It moves now because that mixed graph escaped the workspace.
+  `verifiable-trust-infrastructure` needs 0.19 for two new room specs and could not take
+  it while `affinidi-messaging-sdk 0.22.0` required 0.18: pinning the consumer at 0.19 put
+  two `trust-tasks-rs` nodes in *its* graph and broke `vta-sdk` on exactly that error.
+  Nothing in the consumer's manifests was wrong and no semver check flags it, because the
+  break arrives through a public dependency — the only fix is here.
+
+  0.19 is additive for everything these crates use: it adds two room spec versions and
+  retires their predecessors, and a retired spec keeps its generated module.
+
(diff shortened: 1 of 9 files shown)


stormer78:
Updated after the release guard caught what the first push missed — and one thing the guard cannot catch.

**Version bumps (minor, not patch).** `trust-tasks-rs` is a **public** dependency of `affinidi-messaging-sdk`: `account::update::v0_1::MediatorAcl` and `account::add::v0_1::MediatorAcl` appear directly in its public signatures. So moving it changes these crates' APIs even though not a line of their own source did — a patch bump would hand consumers a silent break on `cargo update`, which is precisely the failure this PR exists to fix. `sdk` 0.23.0, `didcomm-service` 0.7.0, `mediator` 0.23.0, `test-mediator` 0.6.0.

**`affinidi-tdk` 0.12.0 → 0.13.0, which the guard does not flag.** `check-version-bumps.sh` looks at source paths, and only `affinidi-tdk`'s manifest changed — so it passes without it. But the facade is the *only* path by which the consumer reaches the SDK (`verifiable-trust-infrastructure` has no direct `affinidi-messaging-sdk` dependency; it pins `affinidi-tdk = "0.12"`). Leaving it unbumped would publish a fix that never arrives: crates.io's `affinidi-tdk 0.12.0` would go on requiring `messaging-sdk 0.22` and the consumer would still resolve `trust-tasks-rs 0.18`. It re-exports the SDK, so it takes the same public-dependency change and the same minor bump.

Worth noting as a gap in the guard rather than a mistake in it: a manifest-only change to a facade crate is invisible to a source-path check, and a facade is exactly where that matters most.

Guard now passes locally against `origin/main`; `CHANGELOG.md` records all five.

PR #808, opened 2026-09-16
fix(mediator): enforce the per-DID rate limit on authenticated requests

stormer78:
## Problem

#805 found that `limits.did_rate_limit_per_second` / `did_rate_limit_burst` enforce nothing. `DidRateLimiter` is built (`server.rs`), stored in `SharedData`, garbage-collected, and logged as "Per-DID rate limiting enabled", but nothing calls `check`. Verified on `origin/main` at 7abe6b7: the only callers were its own unit tests.

## Fix

**Where it's charged.** Once per authenticated HTTP request, in `authenticate_token`, after the JWT, the session record and the blocked check have passed. Every authenticating route goes through that function: `/inbound` (DIDComm and TSP, via `MaybeSession`), `/outbound`, `/fetch`, `/list/{..}`, `/delete`, `/whoami`, `/oob`, `/admin/status`, and the `/ws` upgrade (both the header and the subprotocol token paths). It has to be charged after validation. Charging before it would let anyone with a forged token naming a victim DID spend that DID's quota.

**Refusal.** The response is a `429` with the same contract as the per-IP limiter: `x-rate-limit-source: mediator`, `Retry-After`, and `{"error":"rate_limited","limiter":"mediator","scope":"did","message":…,"retryAfterSecs":N}`. `limiter` stays `mediator` because the SDK falls back to it as the *source* when a proxy strips the header. The added `scope` member says which limiter refused. The SDK already reads this as `ATMError::HttpStatus` / `is_rate_limited()`; a test checks both the header path and the stripped-header path. The response comes from a new `AuthError::RateLimited { retry_after_secs }` (`AuthError` is `#[non_exhaustive]`).

**`0` = disabled** is unchanged and is still the default.

**Not charged:**
- **The mediator's own DID and the configured `admin_did`.** Throttling the operator's management plane during an incident locks the operator out.
- **Anonymous sessions**, meaning the inter-mediator relay session and the DIDComm v1 anonymous-forward session. They carry no DID and never reach `authenticate_token`. Inter-mediator hops are always posted anonymously (`ForwardingProcessor` holds no session on the peer), so forwarded traffic is never keyed on a peer mediator's DID. Sharing one bucket across all of them would let one busy peer throttle every other peer's forwards. A forward that a *client* submits is charged to that client, since it is the authenticated sender. The per-IP limit still applies to all of these.

**Metrics.** `rate_limited_total` now has a `scope` label: `ip` or `did`. This matches the label its doc comment already described.

**Startup log.** It now says what is metered and what isn't.

## WebSocket: known gap, documented

The `/ws` upgrade is charged, but frames on an established socket are not. A refusal mid-session has no HTTP status to carry it, and DIDComm's problem-report registry has no rate-limit descriptor. Reusing `limits.queue.sender`, the queue-full descriptor, would tell the client the wrong thing. Closing with 1013 would drop a frame the client already believes it sent. Choosing a signal is a protocol decision, so this PR doesn't invent one. The gap is recorded in the CHANGELOG and next to the key in `conf/mediator.toml` (and `docker/test/conf/mediator.toml`).

## Versions

- `affinidi-messaging-mediator` 0.26.1 → **0.26.2** (patch). All changes are additive: a new variant on a `#[non_exhaustive]` enum, plus new methods/const/fn and a `Debug` impl on `DidRateLimiter`. `check` is unchanged. 0.26.1 was never published, but the version guard compares against base.
- `affinidi-messaging-test-mediator` 0.9.0 → **0.9.1** (patch). Adds `TestMediatorBuilder::did_rate_limit(per_second, burst)`.

Behaviour changes worth noting: a DID over quota now gets 429s where it got nothing before, and `rate_limited_total` series now carry a label.

## Tests

- Unit tests (`did_rate_limiter`): disabled at 0, burst then refuse, independent DIDs, `Retry-After` ≥ 1, exempt DIDs, empty hash never charged, and the 429 contract read back by the SDK's `HttpStatusError::from_parts`.
- Integration tests (`test-mediator/tests/did_rate_limit.rs`, against a running mediator):
  - a DID over quota gets a 429 with the contract on `/whoami`, `/inbound`, and the `/ws` upgrade;
  - a second DID has its own bucket;
  - `0` disables the limit;
  - the admin DID is exempt;
  - anonymous relay `/inbound` traffic is never refused with a 429.

Local runs:
- `cargo fmt --all -- --check`: pass
- `RUSTFLAGS="-D warnings --cfg tracing_unstable" cargo clippy --locked --workspace --all-targets -- -D warnings`: pass
- `cargo test -p affinidi-messaging-mediator --lib`: 209 passed
- `cargo test -p affinidi-messaging-test-mediator --tests`: 88 passed
- the same with `--features tsp`: 134 passed
- the same with `--features didcomm-v1`: 95 passed
- `scripts/check-version-bumps.sh`, `check-changelogs.sh`, `check-test-clock-isolation.sh`, `check-workspace-duplicates.sh`: pass
- Redis-backed storage tests were **not** run, because no local Redis was available. The change doesn't touch the store layer.


stormer78:
## Addresses the two CWE-770 code-scanning findings on `did_rate_limiter.rs` (alerts #118, #120)

**Cause.** `DidRateLimiter` keys a `governor` `DashMapStateStore<String>` with one bucket per authenticated DID. `governor` bounds the rate *within* a bucket but never the *number* of buckets, and `spawn_gc` only reclaims fully-replenished buckets on a 60s tick. Between sweeps, an adversary cycling through many distinct authenticated DIDs faster than they replenish grows the map without bound (memory exhaustion).

**Fix (commit ef810d1).** A hard cap on distinct tracked keys:
- `MAX_TRACKED_DIDS = 100_000`. When the store is full and a **new** DID arrives, the limiter attempts a throttled reclaim of replenished buckets and, if still full, **fails that new DID closed** with the existing `429` contract rather than growing the map.
- A DID **already tracked** is never refused for capacity or evicted mid-window; disabled mode (`did_rate_limit_per_second == 0`) has no store and no cap.
- The inline reclaim is throttled to at most one O(tracked) pass per 100ms, so the admission check can't itself become a CPU-exhaustion lever; the periodic 60s sweep remains the unconditional backstop.
- Membership/length are read from a shared `Arc` handle to the same `DashMap` (governor exposes no membership query), via a thin state-store newtype that delegates to it.

**Same class in the shared per-IP limiter — also bounded (not deferred).** `affinidi-rate-limit`'s per-IP `DashMapStateStore<IpAddr>` had the identical unbounded-key class, and is *worse* (keyed on unauthenticated, client-chosen IPs, reachable pre-auth; an IPv6 /64 is one entry per request). It now has the same cap: `MAX_TRACKED_IPS = 250_000`, throttled reclaim, fail-closed on overflow. Bumped 0.1.1 → 0.1.2.

**Mitigating context** (stated, not relied on alone): every per-DID request is JWT-authenticated first, and the per-IP Tower limiter runs before it.

**Tests:** cap enforced / fails closed for a new key; existing keys unaffected at cap; disabled mode ignores the cap; reclaim frees capacity after replenish — for both limiters. `cargo clippy --workspace --all-targets -D warnings`, fmt, the release guards, mediator + test-mediator + rate-limit tests all green (Redis-backed integration tests not run locally; the did-rate-limit integration test uses the in-memory store and passes).

The alerts should clear on the next scan of the new commit; flagging here so they can be dismissed if they don't.

PR #825, opened 2026-09-18
feat(mediator): let a DID empty its own queue in one call

stormer78:
## The gap

Clearing a queue meant paging it — `/list` is capped at 100 **with no cursor**, `/fetch` pages — and deleting by id 100 at a time, every batch separately authenticated.

That is the recovery path a node needs *precisely when it is already being rate-limited*, which is to say precisely when it cannot use it.

Meanwhile `purge_folder` has existed in every store implementation the whole time (`redis`, `fjall`, `memory`) — reachable only as a side effect of **deleting the account**.

## Why it matters, and why the node in trouble can't fix itself

The queue that strands a deployment is usually not on the node that is failing.

A message is held against the **sender's** account until the **recipient** deletes it. So a receiver whose deletes are failing fills the send queue of every peer talking to it — and at `queued_send_messages_hard` (1000 by default) that peer stops being able to send *anything*. It cannot fix that by fixing itself; it needs either the receiver to recover, or a way to empty its own queue.

That is exactly how a DID hosting control plane wedged: its send queue hit the cap because the VTA's deletes were being rate-limited and silently discarded (fixed separately in #824).

`DELETE /purge/{folder}` is how it gets back, without an operator deleting and rebuilding the account.

## Scope and safety

- **Strictly the caller's own queue.** The DID comes from the authenticated session, never from the request. There is deliberately no way to purge another DID's queue — that would be a denial-of-service primitive handed to anyone with `Local` access, and the recovery case doesn't need it.
- **Same `Capability::Local` gate as `/delete`**, since emptying a queue is a superset of deleting from it and must not be the easier of the two to reach.
- **Destructive and unrecoverable** — purged messages are gone, not returned to their senders. On `Inbox` that is undelivered mail addressed to you. Logged at `info` with count and bytes, because that log line is the only record they existed; the SDK's `purge_queue` doc says so twice.

## Testing

The path segment the SDK builds from `Display` and the one axum deserialises via `#[serde(rename_all = "lowercase")]` come from **two different impls**, so nothing but a test ties them together — and a mismatch would be a 404 on the one call a wedged node makes to recover. `a_folders_path_segment_round_trips` covers both folders.

`cargo fmt --all --check` ✅ · `clippy -D warnings` ✅ 0 issues · sdk lib 139 passed · mediator lib 213 passed

## Related

#824 fixes the delete path that caused the wedge in the first place. This is the recovery hatch for when something like it happens again — there is no delivery counter, visibility timeout, dead-letter queue or poison counter anywhere in the mediator, so the only things that release a message today are a client deleting it or the 7-day expiry sweep.


stormer78:
Fixed the `publish dry-run` failure.

**Cause:** the mediator declared `affinidi-messaging-sdk = { path = …, version = "0.26" }`. Under `cargo publish --dry-run` that resolves the *published* 0.26.8, gets past resolution, and then fails to **compile**:

```
error[E0432]: unresolved import `affinidi_messaging_sdk::messages::PurgeQueueResponse`
```

The guard's `awaits_sibling_release` only recognises *resolution* failures (`failed to select a version for the requirement …`), and deliberately so — a compile error against a published sibling is indistinguishable from "this crate uses an API that does not exist", which is exactly the thing the guard should not wave through.

**Fix:** require `version = "0.26.9"` exactly, which is simply true — the purge route returns `PurgeQueueResponse`, and 0.26.9 is what adds it. That turns the failure into an unmet requirement whose local version satisfies it, so the guard classifies it correctly:

```
Waiting on in-release siblings: affinidi-messaging-mediator
All changed publishable crates pass cargo publish --dry-run.
```

I did not touch the guard. Loosening its detection to accept compile failures would have made it stop catching the case it exists for.

PR #826, opened 2026-09-18
feat(delivery)!: give the outbox a way to be emptied

stormer78:
## The gap

`OutboxStore` had **no way to remove anything**. The whole trait was:

```rust
async fn put(&self, entry: OutboxEntry) -> Result<(), OutboxError>;
async fn get(&self, idempotency_key: &str) -> Result<Option<OutboxEntry>, OutboxError>;
async fn due(&self, now_ms: u64) -> Result<Vec<OutboxEntry>, OutboxError>;
async fn awaiting_confirmation(&self) -> Result<Vec<OutboxEntry>, OutboxError> { Ok(Vec::new()) }
```

`put` upserts, state moves `Queued → Sent → Delivered | Unconfirmed | Failed`, and there it stops. So an outbox keyspace grows for the life of the deployment — and `due()` re-reads and re-decodes **every entry ever written**, every tick. The cost of draining rises with everything that has already drained successfully.

A queue that cannot be emptied is not a queue.

## ⚠️ BREAKING

**`remove` is a required trait method.** Every implementor must add it:

- `InMemoryOutboxStore` — done here.
- `vti-common`'s `VtiOutboxStore` — **needs a matching impl** (follow-up in the VTI repo). It is the only other implementor I found.

Required rather than defaulted on purpose: a default would have to be a silent no-op, which is exactly the behaviour being fixed, kept alive under a name that reads like it works.

Minor bump (`0.1.15 → 0.2.0`) rather than patch, because of that. Nothing in this workspace depends on the crate, so there is no cascade.

## Also added (additive)

- **`OutboxStore::terminal_before(cutoff_ms)`** — terminal entries created at or before the cutoff. Defaulted to `none`, because enumerating by age is store-specific and a store that cannot do it efficiently should keep working rather than fail to compile. The cost is that **such a store never reaps**, which its doc says plainly.
- **`reap::{reap_terminal, reap_loop, TERMINAL_RETENTION, ReapReport}`**.

## Two design points

**Why terminal entries are kept for a window rather than removed on settle.** `idempotency_key` is what makes at-least-once retry safe — the receiver drops a duplicate carrying a key it has seen. Delete the record the moment an entry settles and a retry arriving afterwards reads as new work. The window is 7 days, matching the mediator's own `message_expiry_seconds`: a duplicate cannot arrive from a message the mediator has already expired, so keeping it longer protects against nothing.

**Age is measured from `created_at_ms`**, not from when the entry settled — there is no settled-at field and adding one would need a migration for entries already on disk. Creation time is the more conservative basis anyway: it can only make an entry look *older*, so a reap keyed on it never keeps a terminal entry longer than intended.

## Testing

Four tests, and the one that matters most is `unfinished_work_is_never_reaped_however_old`: a `Queued` entry is unfinished work and a `Sent` one is still awaiting evidence, and age does not make either disposable. Eligibility is re-checked inside the reap rather than trusted from the store's filter, so a store that got its own filter wrong cannot make this delete undelivered work on its behalf.

`cargo fmt --all --check` ✅ · `clippy -D warnings` ✅ 0 issues · 65 tests passed

## Context

Found while tracing why a DID hosting control plane's send queue filled. #824 fixes the delete path that caused that; #825 adds the recovery hatch. This one is the unbounded growth underneath, which was real but not what bit us.


stormer78:
Rebased onto main, and reworked in response to the two security findings. **Both were real**, and fixing the first one also removed the reason this PR was breaking.

### 1. TOCTOU (CWE-367) — `outbox.rs`

`reap_terminal` deleted from a snapshot taken by `terminal_before`, and an entry can be re-queued by `put` in between — a retry of an `Unconfirmed` send, say. A general `remove(key)` would then delete **live work on a stale read**, which is losing a message rather than tidying up after one. Re-reading before deleting narrows that window without closing it.

The primitive is now **`remove_if_terminal(key) -> bool`**: the condition is evaluated where the delete happens, under whatever the store uses to make `put` atomic. The reaper never holds an unconditional delete at all.

`an_entry_requeued_after_the_snapshot_is_not_reaped` reproduces the exact interleaving.

### 2. Silent clock fallback — `reap.rs`

`reap_loop` defaulted `now_ms` to 0 on a `SystemTime` error. It fails *closed*, but silently — and the cutoff decides which records get destroyed, so that decision was being made on a bad reading with nothing to show for it. It now logs and skips the pass; the next tick retries.

### Knock-on: no longer breaking, and no longer duplicating

Making removal conditional also made it **defaultable**. `remove_if_terminal` defaults to removing nothing, matching `terminal_before`'s default of listing nothing — a store opts into reaping by implementing both, and one implementing neither is bit-for-bit unchanged.

So this is **0.1.16, not 0.2.0**, which also clears the `release guards` failure:

```
Workspace duplicate guard
  ok     no workspace member is shadowed by a crates.io copy
```

The 0.2.0 bump had left the published `affinidi-messaging-delivery 0.1.15` in the graph via **`vta-sdk 0.43.0`, which pins `^0.1`**. The honest alternative was an allow-list entry — and that file makes clear those are debt markers this repo has paid for before, once by freezing its own API against the registry. Better not to take on a new one for a change that did not need to be breaking.

I kept the argument I made originally, that a defaulted no-op can hide a broken store — but made it observable instead of relying on the type system: `ReapReport` gains `skipped`, counting entries the store offered and then declined to remove. `terminal_before` implemented without `remove_if_terminal` is the one incoherent pair, and it now shows up as a count equal to the offered total on every pass.

Also drops the follow-up this PR was creating: `VtiOutboxStore` no longer *has* to change, and can opt into reaping when VTI chooses.

`clippy -D warnings` 0 issues · 68 tests passed (7 in `reap`, 3 new).

PR #829, opened 2026-09-19
fix(mediator)!: run the queue gates on direct delivery too

stormer78:
Follows #828. **It also withdraws two claims made in that PR's body** — see the last section.

## The gap

The three queue-depth gates — per-relationship, sender total, recipient total — live in `protocols::routing`, which exists only in a `didcomm` build and is reached only by a `forward`.

Direct delivery (`inbound.rs:1231` — a client handing the mediator an already-packed envelope for a local DID) and the TSP bridge (`inbound.rs:674`) go through `messages::store::store_message`, which validates fan-out width (`to_recipients`) and message expiry and **no queue depth at all**.

So the per-relationship gate added in #828 never protected that path. A sender using direct delivery could fill any inbox without bound, with nothing but the expiry sweeper to stop it.

A unit test on the gate would not have caught this, because the gate was correct — it was simply never called here.

## The change

The gates move to `messages::queue_limits`, outside the `didcomm` gate, and identify a message by id rather than by a parsed `Message`, which the direct path does not have. `validate_store_queue_limits` fetches the two account records and runs all three.

The forward path keeps calling the three directly: it has already resolved both accounts, and re-reading the sender there would re-run the auto-registration that resolution performs.

**Mediator-generated replies stay exempt**, and that is load-bearing rather than a convenience. They are how a client is told its queue is full — gating them turns a full inbox into an *unreportable* inbox, leaving the peer with a silence it cannot distinguish from a dead mediator. The exemption falls out of where the check sits: the `Envelope` arm is client traffic, the `Message` arm is the mediator's own (its sender is `mediator_did_hash`).

## This is a real behaviour change, and a test proved it

`a_slow_consumer_does_not_lose_packed_frames` pushes 130 TSP frames from one sender to one idle recipient, to prove the pickup socket discards nothing. That is by construction the shape the gate refuses — it failed at frame 49, exactly as `queued + incoming >= limit` predicts for a cap of 50.

The behaviour is correct: a sender holding 50 undelivered messages for a peer that is not collecting now gets a typed 503 instead of unbounded growth. The test disables the gate explicitly, so `limits.queue.peer` stays covered by the new file and that one stays about frames surviving a slow consumer.

**Operators should expect the same:** bulk delivery to a peer that is not collecting is now refused at the per-relationship cap rather than queued without limit. That is the point of the change, but it is a change.

## Tests

New `direct_delivery_queue_limits.rs` drives the real protocol end to end — a unit test could not have found the original gap, so these assert on the recipient's mailbox as well as the send result:

- refused at the per-relationship cap, and the refused message is **not** stored;
- a full relationship does not block a different recipient (the #828 fan-out case, now on the direct path);
- collecting reopens a capped relationship;
- `-1` disables the gate without disabling the other two.

They name the mediator's existing `queued + incoming >= limit` convention in a `FITS` constant rather than encoding the off-by-one as a bare expectation.

`TestMediatorBuilder::queue_send_limit_per_peer` is added so a test can reach the refusal in a few sends rather than fifty.

## Verification

- mediator lib: 291 pass
- test-mediator: all four CI variants (memory, fjall, tsp, didcomm-v1) green
- `clippy --all-targets -- -D warnings` clean on the fjall/memory build **and** a `redis-backend,didcomm` build
- `scripts/check-version-bumps.sh origin/main` — both crates `ok`

## Withdrawing two claims from #828

Both were in that PR's body and both were wrong:

1. **`delete_folder_stream` does not drift the account counters.** Its two callers either route through `delete_message` (which decrements) — `purge_folder` — or delete the account record outright — `account_remove`. The leftovers `purge_folder` sweeps are messages whose `delete_message` *failed*, which still exist and are still legitimately counted, so not decrementing is correct. I read the method in isolation and did not check its callers. It is on the public trait, so an external implementor calling it directly could drift; that is a doc hazard, not a bug.

2. **The receive-side mirror was already bounded.** `queued_send_messages_per_peer` (50) sits below `queued_receive_messages_soft` (200) and both key on the same `(from, to)` pair, so one sender could already occupy at most 25% of a recipient's allowance. A second per-pair gate on the receive side would read the same number for the same pair.

One residual from (2) is **not** addressed here: nothing enforces `per_peer < receive limit`. An operator setting `per_peer = -1` or raising it restores monopolisation, and `queue_receive_limit` is settable per account, so an account with a custom limit below 50 is monopolisable on today's defaults. Worth its own change.


albertoleon7794:
Ran our vetting ceremony against this branch (`b544da04`) with the rest of the stack on main. It passes — but only after clearing a pre-existing queue by hand, and the reason is worth a look now that this has landed.

The DID-hosting daemon's outbound queue held 223 messages for a single VTA, oldest 2.6 days. 57 were still in the recipient's inbox; **166 had already been delivered** — the recipient drained them, classified them as terminal errors, and dropped them without answering. That is the correct behaviour after the error-loop fix. But a drop produces no acknowledgement, so all 223 stayed in the sender's queue. Now that the depth gates run on the direct-delivery path, the mediator refuses everything that daemon sends with `e.p.limits.queue.sender` — including traffic that would otherwise succeed — so DID creation stops for every client behind it. The queues did not drain on their own across repeated checks; the expiry sweeper had not touched a 2.6-day-old queue.

The general shape: any message a recipient deliberately declines to process — terminal error, unsupported type, policy refusal — is delivered, never acknowledged, and counted against the sender indefinitely.

Two ways out, both yours to pick: treat the acknowledgement as "stop holding this for me" rather than "I acted on it", so a terminal drop still acks; or extend the eviction path to cover delivered-but-unacknowledged messages, not only undeliverable ones.

Nothing against this change — the gates are right and we are glad they landed. Flagging it because this is what turns a stale queue into an outage for the sender. Details and raw counts: https://github.com/berkmancenter/keyring-wallet/blob/26104cea58763735292f2f70093ae8b1c4c4324d/docs/VTI_UPSTREAM_FINDINGS.md#vti-31--a-dropped-terminal-error-is-never-acknowledged-so-it-never-leaves-the-senders-queue. Happy to open a separate issue to track it, since this one is already merged.

**TL;DR — we tested this PR and it works. One heads-up: messages a recipient throws away are never acked, so they pile up in the sender's queue forever. The new limit counts them, so a sender can get blocked over messages that were already delivered. Worth a separate fix.**

PR #830, opened 2026-09-19
fix(mediator): tell a client when its live notification was dropped

stormer78:
Closes KR-30 (their VTI-30).

## The bug is the silence, not the drop

A live notification is dropped when the client's send queue is full (`WS_CHANNEL_SLOTS` = 8) or the global byte budget is exhausted. Nothing durable is lost — the message stays in the recipient's inbox — and the drop was **already** counted as `ws_live_delivery_dropped_total`.

What was missing is that nothing was said on the wire. A client that only *listens* never learned there was anything to collect, so it did not collect it. The existing log even says "Client will pick the message up from its inbox" — which is true, and which the client has no way of knowing.

## The fix

The streaming task raises a per-connection resync flag when a push cannot be queued. The socket handler reads it after every wake-up and sends a message-pickup 3.0 `status` carrying the live `message_count`, as soon as the socket is moving again.

A client already understands that message — it is exactly what a `status-request` returns — so recovery needs **no new protocol on the client side**.

In TSP mode a notification is only ever a wake-up (the body is ignored and the handler re-drains), so there the resync *is* the drain and no separate signal is sent.

## Why a flag rather than another frame

The drop happens precisely when there is no room to queue anything, so a resync frame pushed through the same bounded channel would be dropped by the same congestion. The flag is generated at the writer, which owns the socket, so it bypasses the queue entirely.

Repeated drops collapse into one signal. The client's answer to any number of them is the same single drain, and a congested socket is the worst possible place to add traffic. The handler takes the signal with `swap`, so a drop landing *during* the send raises it again rather than being folded into one already on its way out.

## `deliver_live`

Queueing and flag-raising moved into one helper. They are a single decision, and splitting them is how a future caller ends up doing the first without the second — which is the exact shape of this bug. It also makes the behaviour testable: the delivery arm was inline in the select loop and could not be driven by a test.

## Tests

- `a_dropped_live_notification_raises_resync` — a push into a full channel raises the flag; a delivered one does not.
- `repeated_drops_collapse_into_one_signal` — five drops leave one pending signal, not five.

**Coverage limit, stated plainly:** these cover the *raise* side. The send side — that the packed `status` actually reaches a congested client — is not covered end to end, because provoking a full send queue through the harness needs a client that connects and then refuses to read, which the SDK client does not do. Reviewer attention there is worth more than on the rest of this.

## New metric

`ws_live_resync_sent_total`. It deliberately does not match `ws_live_delivery_dropped_total` one for one, since drops coalesce. **Drops climbing while resyncs stay flat is the alertable shape**, not the drops themselves.

## Verification

- mediator lib: 293 pass (2 new)
- test-mediator: all four CI variants green (memory, fjall, tsp, didcomm-v1)
- `clippy --all-targets -- -D warnings` clean on fjall/memory **and** `redis-backend,didcomm`
- `scripts/check-version-bumps.sh origin/main` — `ok`

Additive: no configuration or API change.


stormer78:
Both AgenticSec findings reviewed. **One taken, one rebutted.**

## Alert 83 — unbounded resync packing (HIGH, confirmed) — fixed in 900bf32

This one is fair and I've taken it. I added an attacker-influenceable crypto path with no throttle: each resync costs an inbox read, a DID resolution and a `pack_encrypted`, and the party deciding how often it runs is the congested client, by choosing how slowly to read.

A 5s per-socket floor now bounds it (`MIN_RESYNC_INTERVAL`). Worth saying that the floor is also correct with no attacker involved, for the same reason the drops already coalesce: the client's answer to any number of dropped notifications is one drain, so a second `status` before it has acted on the first tells it nothing — on the one socket already known to be congested.

**The suggested remediation has a bug, and it's the one that matters here.** The proposed code does:

```rust
let due = live_resync.swap(false, Ordering::Relaxed)
    && now.duration_since(last_resync_sent) >= MIN_RESYNC_INTERVAL;
```

`swap` clears the flag unconditionally. When the interval has *not* elapsed, the signal is destroyed by the very branch meant to defer it, and the trailing `else if live_resync.load(...)` reads the `false` that swap just wrote. That converts a rate-limit into silent signal loss — the exact failure this PR exists to fix.

The flag is therefore **read**, not swapped, until the floor has elapsed, and only cleared immediately before a send that is actually happening.

I also took the recommendation for a second counter. Without it the alerting guidance in the previous commit becomes wrong: with a floor, a flat `ws_live_resync_sent_total` beside climbing drops is *correct* under sustained congestion. `ws_live_resync_suppressed_total` is what separates suppressed-by-design from never-generated.

`resync_due` is a free function so the rule is unit-testable — the socket loop cannot be driven from a test, and "is it raised" and "has the floor elapsed" are precisely the pair that is easy to get wrong together.

## Alert 151 — resync flag orphaned during duplicate-DID churn (MEDIUM, must-review) — not a defect

Rebutting this one, with evidence.

The scanner's own validation log says the blocking gap was that "the code path for what happens to a refused registration's socket ... is not shown in full". It is shown, in a test that predates this PR: **`duplicate_replacement_redelivers_inbox_to_new_socket`** (`websocket_streaming.rs`).

On duplicate-DID replacement the displaced socket is sent `Close(CloseReason::Replaced)` and **the replacing socket is redelivered the inbox** — the test asserts it receives the stranded message body. So a resync flag orphaned on the displaced socket's `Arc` costs nothing: the new socket is handed the actual messages, which is strictly more than the signal would have told it. Propagating the flag across the churn would ask the new socket to go looking for something it has already been given.

For the refused-registration sub-case: a socket refused by the duel damper never owns the slot, so it is not the one `deliver_live` signals, and it is closed rather than left listening. There is no client left waiting on that `Arc`.

Severity was self-assessed as "a race-dependent UX nuisance" with "dropped messages remain durably queryable", which is the right read — and the redelivery path means even the nuisance does not occur.

Happy to add an explicit assertion on `.resync` state after churn if a reviewer would rather have it pinned than argued, but I don't think the behaviour needs changing.

## Verification after the fix

- mediator lib: 294 pass (1 new: `resync_is_due_only_when_raised_and_past_the_floor`)
- `clippy --all-targets -- -D warnings` clean on fjall/memory and `redis-backend,didcomm`

PR #831, opened 2026-09-19
docs(mediator): state the Origin rule by the header, not the client class

stormer78:
Closes KR-05 (their VTI-05). **No behaviour change** — the `Origin` check is untouched and the default stays closed. What changes is that the documentation no longer contradicts it.

## The defect was a true sentence

Three places said the same thing:

- `conf/mediator.toml` — *"header-less native clients are unaffected"*
- the setup wizard's CORS screen (`mediator-setup/src/app.rs`) — *"Native clients (no Origin) are unaffected"*
- this crate's own comment above the check (`handlers/websocket.rs`) — *"Native clients send no Origin and pass straight through"*

Every one of those sentences is true. Together they mislead, because a reader substitutes **"native"** for **"sends no Origin"**, and those are not the same set.

**React Native's WebSocket sends an `Origin` derived from the connection URL.** A mobile wallet is therefore refused by the default policy exactly as a browser is:

```
WebSocket upgrade rejected: Origin not permitted by CORS policy
```

An integrator hitting that had three of our own documents telling them it could not be happening. The Rust SDK sends no Origin, which is why first-party testing never met it.

## The rule, as it now reads everywhere

> A request sending **no** `Origin` is admitted.
> A request sending one is admitted only if the policy admits that origin.

The header is the distinction, not the client class. Same rule on REST and on the WebSocket upgrade.

## Test names were the fourth copy

`origin_check_allows_header_less_native_clients` and `origin_check_none_policy_rejects_browser_origins` passed the whole time while naming the wrong distinction — a test that passes while documenting the wrong idea is part of how the idea survives. Renamed to `origin_check_admits_a_request_with_no_origin_header` and `..._refuses_any_origin_header`, and added `origin_check_refuses_a_react_native_client_under_the_default`, which pins the case that was missed and asserts it is admitted once allowlisted.

## Also

- New `docs/cors-and-origin.md`: the rule, the three policies, how to read the rejected origin out of the mediator's own log, and why the default stays closed.
- The refusal log now names `[security] cors_allow_origin` and that doc. **The response body to the caller is unchanged**, so nothing new is disclosed to an unauthenticated peer — the detail goes to the operator, who is the one who can act on it.

## Why the default is not changed

A wildcard would be safe on this service — every endpoint takes a bearer token in a header, never an ambient cookie, so `"*"` creates no CSRF exposure. It is still not the right default: an operator who has not considered browser access should not silently have it, and a default-open posture that turns out wrong is not recoverable by configuration, whereas this one costs a single config line plus the new page.

The default was never the defect. The claim that it could not affect a native client was.

## KR-06 does not reproduce

Their VTI-06 says `cors_allow_origin` is ignored unless it sits in `[security]`. Against this tree it does not reproduce: the reference config documents it under `[security]` (`conf/mediator.toml:241`), the wizard writes it there (`config_writer.rs`), and the parser reads it from there. It needs their `mediator.toml` to go further, so it stays open pending that.

## Verification

- mediator lib: 292 pass (1 new, 2 renamed)
- setup wizard: 407 pass
- `clippy --all-targets -- -D warnings` clean on both crates
- `scripts/check-version-bumps.sh origin/main` — `ok`. The setup tool is `publish = false` and takes no bump.

## Note for whoever merges

This and #830 both bump the mediator to **0.28.1**. Whichever lands second needs a rebase and a re-bump to 0.28.2 — they touch different files, so the only conflict is the version line and the changelog.


stormer78:
Rebased onto `main` now that #830 has landed.

The only conflict was the changelog: #830 took `0.28.1`, so this is now **0.28.2** and its section sits above #830's. `Cargo.toml` dropped out of the rebase entirely — my bump to 0.28.1 had become identical to the one already on main — so I re-bumped it deliberately rather than let the branch ship a version that is already spoken for.

`handlers/websocket.rs` auto-merged cleanly: #830's changes are in the resync block and the metrics import, this PR's are in the Origin comment, the refusal log and the test names. Verified both survive — the resync debounce (`MIN_RESYNC_INTERVAL`, `resync_due`, `WS_LIVE_RESYNC_SUPPRESSED`) and the renamed origin tests are all present and passing together.

Re-verified after the rebase: 295 mediator lib tests pass, `clippy --all-targets -- -D warnings` clean on the mediator and the setup wizard, `cargo fmt --all --check` clean, version guard `ok` (0.28.1 → 0.28.2).

PR #855, opened 2026-09-21
feat: a live traffic monitor for the mediator

stormer78:
This PR serves the `messaging/monitor/*` tasks (trustoverip/dtgwg-trust-tasks-tf#549): a **live traffic monitor** for troubleshooting a mediator. It shows what is coming in and going out, over DIDComm or TSP, what was refused and why, and where each message went.

### What a subscriber sees
`messaging/monitor/subscribe` opens a leased, filtered tap. The lease defaults to 5 minutes, is at most 1 hour, and can be renewed or re-filtered by id. The mediator pushes **signed** `messaging/monitor/event` batches over the subscriber's **live connection only**. Each event carries metadata only, **never a message body**.

| Stage | Where it's observed |
|---|---|
| `received` | all six inbound entry points: REST and websocket × DIDComm v2, DIDComm v1, TSP |
| `refused` | the same entry points, with the problem-report code and comment |
| `stored` | direct delivery and Routing-2.0 local forwards |
| `delivered` | live push at store time, and REST `/fetch` |
| `forwarded` | DIDComm and TSP forwards enqueued for another mediator |
| `deleted` | REST `/delete` and `messaging/message/delete` |
| `purged` | REST `/purge` and `messaging/queue/purge` |

Every event also carries the sender, recipient, size, message id, channel and protocol. **Filters** are ANDed together, and values within one filter are ORed. Available filters: DIDs (matching either party), direction, stage, protocol, channel, message-type prefix, and failures only.

**Scope.** An administrator may watch everything. Any other account sees only its own traffic: if the `dids` filter is omitted it is narrowed to the account itself, and naming another account is **refused** rather than silently narrowed. Each account may hold at most three subscriptions. An unsubscribe of someone else's id gets the same answer as an unknown id.

### A monitor must not hurt what it watches
- **Free when idle.** `TrafficMonitor::emit` takes a closure and builds the event only if a subscription exists. An unwatched mediator pays one atomic load per hook.
- **Never queued or stored.** Events that exceed the rate ceiling, that a lagging receiver misses, or that arrive while the subscriber is offline are **counted** in `dropped` and not buffered. A monitor can't fill the queue it observes or slow down delivery.
- **Doesn't watch itself.** Batches go straight to the streaming publisher, which emits no events. By default the subscriber's own management traffic with the mediator is also left out, since console polling would otherwise dominate the feed.
- **Gaps and dead taps are visible.** `seq` goes up by 1 for each batch actually handed to the connection, so a gap means a batch was lost in transit. An empty heartbeat batch is sent every 30 seconds, so a dead tap shows up as dead rather than quiet (R6.2).

### SDK
- `trust_tasks().monitor_subscribe(profile, filter, lease_seconds, max_events_per_second, renew)`
- `trust_tasks().monitor_unsubscribe(profile, id)`
- `protocols::trust_tasks::decode_monitor_event(&Message)`: picks monitor batches out of the profile's live stream.

### Not yet observed (a follow-up PR)
- **Forwarding-processor outcomes** (relayed, failed, abandoned). The processor lives in `mediator-common` with no `SharedData`, so it needs an injected sink, the way `with_system_packer` is injected.
- **Per-message expiry.** `ExpiryReport` only carries counts, so this needs a store-trait change.
- **Deliveries on DIDComm pickup, TSP socket drain, and websocket redelivery.**
- **Subscribers on a raw-TSP socket.** Those sockets treat a push as a wake-up signal and discard the body.
- **Multiple instances.** Behind a shared Redis, a subscriber sees only the traffic of the instance its socket is on.

### Versions
- `affinidi-messaging-mediator` 0.28.19 → 0.28.20
- `affinidi-messaging-sdk` 0.26.17 → 0.26.18
- `affinidi-messaging-test-mediator` 0.9.9 → 0.9.10 (tests only)

CHANGELOG entries and `Cargo.lock` are included.

### Verification
- **Unit tests (`monitor::tests`):**
  - filter semantics: an empty filter matches all, DIDs match either party, ANDed members and ORed values, failures only, message-type prefixes
  - with no subscriber the event is never built
  - event JSON carries metadata only
- **End-to-end tests (`mediator_ops`, 14/14):**
  - An admin subscribes to bob's traffic, alice messages bob, and the admin receives a **signed** batch on its websocket. The batch shows `stored` from alice to bob over `didcomm`, with no body. `monitor_unsubscribe` then reports events sent.
  - A standard account can't monitor another account. Its own subscription is narrowed to itself, and another account can't end it.
- **Whole `affinidi-messaging-test-mediator` suite:** all pass, including the data-plane suites the hooks touch.
- mediator lib (`tsp`): 293 passed
- `cargo clippy --all-targets -D warnings`: clean on `tsp,didcomm-v1` and on `didcomm,memory-backend`. The `tsp`-only (no `didcomm`) combination already fails to build on `main`, in code this PR doesn't touch.
- `scripts/check-version-bumps.sh`: ok

### Pre-merge checklist (vti-stack-development-guide §9)
- [x] No new reqwest::Client::new() / bare fetch(); all clients have finite timeouts (R1.2)
- [x] No lock held across a network await (R1.3): a subscription's settings mutex is released before the store lookup, packing, and publishing
- [x] No local state committed before its remote effect, or the flow is resumable with an idempotency key (R2.1): the monitor holds no durable state. Subscriptions are in memory and end at lease expiry or restart.
- [x] Every retry is bounded + backed off; non-idempotent ops are not blind-retried (R1.4): nothing is retried. A batch that can't be delivered is counted as dropped.
- [x] Accept/poll/listen loops survive transient errors (R1.5): a broadcast lag is counted, and pack or publish failures are counted and logged at debug. The loop ends only on cancel, lease expiry, or bus close.
- [x] Acks/deletes happen only after durable handoff (R1.6): n/a
- [x] New/changed wire types: camelCase, deny_unknown_fields where security-relevant, schema registered, all consumers (incl. JS) updated (R3.*): specs in #549
- [x] Config absence = most restrictive; fail-closed if enforcement can't start (R5.*): non-admins are scoped to themselves, and cross-account requests are refused
- [x] Logs/status claim only what was verified; background-job failures are surfaced (R6.*): `dropped`, `seq` gaps and heartbeats make loss and dead taps observable. `delivered` at store time means "handed to the live stream", which is what the streaming task does.
- [x] "Process dies on the next line" answered for every mutation touched (R2.1): no data-plane behaviour changes, only observation
- [x] Deviations from this guide flagged explicitly with rule numbers: none


stormer78:
Security review follow-up in 9d0db2e2:

- **Code-scanning alert 218:** fixed. Every administrator monitor subscribe, renew and unsubscribe is now audited (`monitorSubscribe` / `monitorUnsubscribe`, in `affinidi-messaging-mediator-common` 0.16.15). The `mediator_ops` e2e test asserts both entries.
- **Code-scanning alert 216:** fixed. A refusal that isn't a problem report is now reported to subscribers as `internalError` with no detail. A unit test covers it.
- **Code-scanning alert 220:** by design. A non-admin subscription is confined to events in which that account is a party. The `monitor` module docs now state what a subscription reveals. I've left dismissing the alert to a maintainer.

